PR #163, opened 2015-01-04
Final state of MP Toolbox

LostSign:
Final version of matching-pursuit-toolbox for design project.


chdinh:
Checked. Good Job.
(Note: Check utils.pro if it still compiles on HP-UX with the new Qt includes network + concurrent)

PR #166, opened 2015-02-05
Remove TRG_xxx channel from the channel name init.

liminsun:


liminsun:
Remove TRG_XXX channels at SQUID Control DLG

PR #171, opened 2015-06-15
Change the status of projectors to inactive in order that MNE can pro…

liminsun:
…cess the data properly.


liminsun:
The active projector will cause the data un-modifiable.

PR #172, opened 2015-07-06
MNE Bem Libs

JanaKiesel:
Add MNEBem and MNEBemSurface class to mne Lib. 
Add readBem example, that loads Bem data from a fif file.


LorenzE:
Thx Jana. Great work!

PR #177, opened 2015-12-13
Write matrix as row-major

chdinh:
Fix int and float matrix writing


MNE/rtClient/rtdataclient.cpp
```diff
@@ -290,7 +291,8 @@ FiffInfo::SPtr RtDataClient::readInfo()
                                 p_pFiffInfo->comps[countComp].data->col_names = FiffStream::split_name_list(col_names);
                             break;
                         case FIFF_MNE_CTF_COMP_DATA:
-                                p_pFiffInfo->comps[countComp].data->data = t_pTag->toNamedMatrix().cast<double>();
+                            //ToDo: Test; Float Matrix
+                            p_pFiffInfo->comps[countComp].data->data = t_pTag->toNamedMatrix().cast<double>();
```
LorenzE: Was this tested using the FiffSimulator and a fiff file with available compensators (i.e. a ctf file)?

MNE/rtClient/rtdataclient.cpp
```diff
@@ -290,7 +291,8 @@ FiffInfo::SPtr RtDataClient::readInfo()
...
                             break;
                         }
                     }
```
LorenzE: Are these two classes (rtdataclient and mne_surface) really the only classes which use toFloatMatrix()? Just want to double check here :)

LorenzE:
Nicely done!

PR #244, opened 2016-12-08
Dipole fit

chdinh:


chdinh:
Dipole Fit is a Technical Preview (TP), which already includes the complete port of all MNE-C dipole fit routines. Anyhow, still a lot of refactoring needs to be carried out to meet the MNE-CPP coding conventions.

PR #260, opened 2017-02-17
Fix buffer problem EEGoSports plugin

LorenzE:
Fix problem with an overflowing device buffer in eegosports plugin. Also, refactor plugin to meet coding conventions. This is solution is still untested -> WIP


chdinh:
Nice work :-)!

PR #263, opened 2017-02-28
[LIB-156] Refactoring

chdinh:


examples/ex_connectivity_tutorial/ex_connectivity_tutorial.pro
```diff
@@ -1,6 +1,6 @@
 #--------------------------------------------------------------------------------------------------------------
 #
-# @file     connectivityTutorial.pro
+# @file     ex_connectivity_tutorial.pro
```
LorenzE: do we want to have tutorial in here? example or tutorial we should decide for one. I like ex_ a bit better.

examples/ex_connectivity_tutorial/ex_connectivity_tutorial.pro
```diff
@@ -44,7 +44,7 @@ QT += widgets 3dextras
 CONFIG   += console
 CONFIG   -= app_bundle
 
-TARGET = connectivityTutorial
+TARGET = ex_connectivity_tutorial
```
LorenzE: see above

PR #272, opened 2017-03-09
Fix dipole moment visualization

LorenzE:
- Fix problem in FiffCoordTrans and Disp3D


MNE/disp3D/model/items/sourceactivity/ecddatatreeitem.cpp
```diff
@@ -204,23 +204,26 @@ void EcdDataTreeItem::setVisible(bool state)
 void EcdDataTreeItem::plotDipoles(QSharedPointer<ECDSet> pECDSet)
```
chdinh: How about no shared pointer here - just a const reference would be great

MNE/disp3D/model/items/sourceactivity/ecddatatreeitem.cpp
```diff
@@ -204,23 +204,26 @@ void EcdDataTreeItem::setVisible(bool state)
...
     for(int i = 0; i < pECDSet->size(); ++i) {
         pos.setX((*pECDSet)[i].rd(0));
         pos.setY((*pECDSet)[i].rd(1));
         pos.setZ((*pECDSet)[i].rd(2));
 
-        to.setX((*pECDSet)[i].Q(0));
-        to.setY((*pECDSet)[i].Q(1));
-        to.setZ((*pECDSet)[i].Q(2));
-        //to.normalize();
+        norm = sqrt(pow((*pECDSet)[i].Q(0),2)+pow((*pECDSet)[i].Q(1),2)+pow((*pECDSet)[i].Q(2),2));
```
chdinh: could you use the norm to scale the arrow up and down - like 1 nAm equals 1cm or something like this

PR #287, opened 2017-04-07
Uncommented important error message

er06645810:
I noticed this extremely helpful error was commented out. Without this print statement, `push` fails silently, and it can take a long time for a developer new to the mne-cpp ecosystem to dig down far enough to find the problem.

Is there a reason it was commented out in the first place?


LorenzE:
@er06645810 Thx for pointing this out. Apparently I commented out this section years ago, when I added the sempahore to the buffer.

PR #291, opened 2017-04-22
ENH: Add extra fields to IO roundtrip

### LINKED ISSUE #280: BUG: Fields discarded by mne_browse I/O roundtrip
Doing a save/load in `mne_browse`, I get a number of fields discarded. Using `sample_audvis_raw.fif` I see these missing according to `mne_show_fiffd`:
```
       212 = scientist
       206 = comment
       104 = {	115 = events
          600 = event ch #'s
          601 = event list
       105 = }	115 = events
       500 = project id
       501 = proj. name
```
It's possible that other fields that are not present here, but are present in general (e.g., subject info, processing history, etc.) are also not read/written. These are detailed in [read_meas_info](https://github.com/mne-tools/mne-python/blob/master/mne/io/meas_info.py#L606) in Python.

Eric89GXL:
Closes #280.

The best solution for the "values" part would probably be to make `info` actually contain the values, but in the meantime this is a workaround.


chdinh:
Great, thank you!

chdinh:
Looks good to me.

PR #292, opened 2017-04-24
[WIP] Improve HPI thread handling

LorenzE:
- Cleanup HPI fitting code.
- Improve QThread handling for rtHPI (based on QThread docu).
- This might fix problems with babyMEG - Still needs to be tested.


MNE/rtProcessing/rthpis.h
```diff
@@ -104,14 +104,42 @@ struct FittingResult {
...
+{
+    Q_OBJECT
+
+//=============================================================================================================
+/**
+* Real-time HPI worker.
+*
+* @brief Real-time HPI worker.
+*/
+public slots:
```
chdinh: if you use new signal slot syntax - slots do not have to be declared anymore:
public: instead of public slots:

MNE/rtProcessing/rthpis.cpp
```diff
@@ -79,113 +79,93 @@ using namespace INVERSELIB;
...
+                    vFreqs,
+                    fitResult.errorDistances,
+                    fitResult.fittedCoils,
+                    pFiffInfo);
+
+    emit resultReady(fitResult);
 }
 
 
 //*************************************************************************************************************
```
chdinh: Consider putting a more explicit command line here  - since a different class starts below:

 // DEFINE MEMBER METHODS OF RtHPIS

MNE/rtProcessing/rthpis.cpp
```diff
@@ -79,113 +79,93 @@ using namespace INVERSELIB;
 // DEFINE MEMBER METHODS
```
chdinh: Since there are residing two classes in this file consider having a more explain comment here:

============================================================================================================= 
 // DEFINE MEMBER METHODS OF RtHPISWorker

PR #296, opened 2017-04-29
MNE Analyze first Extensions ready

chdinh:


MNE/deep/deep.cpp
```diff
@@ -288,7 +288,7 @@ bool Deep::loadModel(const QString& modelFileName, const DeviceDescriptor &devic
 
     fprintf(stderr, "Loading model %s.\n",modelFileName.toUtf8().constData());
 
-    m_pModel = Function::LoadModel(modelFileName.toStdWString(), device);
```
LorenzE: mind naming conventions. loadModel(...) . same for below

chdinh: CNTK function

PR #298, opened 2017-05-19
Added Fedora library extensions to gitignore

er06645810:
Compiling in Fedora produces `.debug` and `.so.debug` files in the /bin directory. This commit adds such files to .gitignore so that they are not tracked.


LorenzE:
Thx @er06645810

PR #304, opened 2017-07-11
Minor improvements

LorenzE:
- Add alpha material usage as default in Disp3D
- Fix model rotation problem in Disp3D
- Fix connectivity thresholding in Disp3D
- Remove SCDC recalculation when updating bad channels


MNE/disp3D/engine/model/items/network/networktreeitem.cpp
```diff
@@ -238,7 +238,7 @@ void NetworkTreeItem::plotNetwork(const Network& tNetworkData, const QVector3D&
 
         //Plot out edges
         for(int j = 0; j < lNetworkNodes.at(i)->getEdgesOut().size(); ++j) {
-            if(abs(lNetworkNodes.at(i)->getEdgesOut().at(j)->getWeight()) >= vecThreshold.x()) {
+            if(fabs(lNetworkNodes.at(i)->getEdgesOut().at(j)->getWeight()) >= vecThreshold.x()) {
```
1DIce: Use std::fabs or std::abs instead, because using abs() or fabs() calls the old C functions.  abs() is not overloaded for float types, std::abs() is. So everytime abs(3.0) is called its gets converted to abs(3). This error is multiple times in this file and probably in others aswell.

LorenzE: Thx for the hint. fabs() without std:: should work fine here though. But for the sake of good practice, I changed it to use the namespace std:: and to prevent wrong abs() usage (ex. for floating values) in the future. I double checked some code and from what I saw fabs() or abs() are correctly used when dealing with floating or integers values. Please let me know if you find otherwise.

1DIce: I found about 20 times where abs() is used for floats. I can go through and change them after this PR if u want.

LorenzE: Awesome! That would be great!

PR #307, opened 2017-08-30
Changed = to == inside if statement

er06645810:
There is an `if` statement in `epidetect.cpp` on line 347 that appears to be missing an `=`. I believe it is meant to be checking the value of `counter`, but instead it is overwriting it.


LorenzE:
thx @er06645810

PR #310, opened 2017-09-04
Lib refactor

LorenzE:
Refactor MNE-CPP lib structure.


MNE/disp3D/helpers/interpolation/interpolation.h
```diff
@@ -33,16 +33,16 @@
 *
 */
 
-#ifndef INTERPOLATION_H
-#define INTERPOLATION_H
+#ifndef DISP3D_H
```
1DIce: This should probably stay INTERPOLATION_H

chdinh:
Good job guys!

PR #312, opened 2017-09-27
New buffer for both workers

1DIce:
Created a new class DynCircularBuffer. This new class is now used in both workers as queue for the real time data and replaces the QList. 
The DynCircularBuffer is an array based queue with dynamic size and will hopefully perform better than QList.
I also changed View3D's default setting to render new frames only if needed.


1DIce:
Sadly i had to remove the changes from View3D, because the function was introduced in Qt5.9.

libraries/utils/generics/dyncircularbuffer.h
```diff
@@ -0,0 +1,521 @@
...
+
+    //Set tail to the front of the buffer if needed
+    if(m_iTail == m_iBufferSize)
+    {
+        m_iTail = 0;
+    }
+
+    if(m_iTail == m_iHead)
+    {
+        reserve(2 * m_iBufferSize);
```
chdinh: Please check,, produces a memory leak

libraries/utils/generics/dyncircularbuffer.h
```diff
@@ -0,0 +1,521 @@
...
+
+
+//=============================================================================================================
+/**
+* This class is a implementation of a queue based on a array with dynamic size.
+*
+* @brief Dynamic circular queue/buffer.
+*/
+template<typename T>
+class DynCircularBuffer
```
chdinh: Dynamic circular buffer is not really useful. you could use a queue instead

libraries/utils/generics/dyncircularbuffer.h
chdinh: produces leaks - what happens to the data stored in the old buffer?

libraries/utils/generics/dyncircularbuffer.h
1DIce: It is a queue, but array-based for faster access and the [ ] operator can be used. You are right i should rename it.

chdinh: And why has it to be circular than?

1DIce: It minimises the number of  memory allocations, because space that is set free by pop_front() can later be used by push_back(). Maybe this is helpful:
arrayqueue

chdinh: I see, you basically safe the request and allocation of new memory from the heap.

1DIce: Yes allocation is only needed when all slots in the array are full, when this happens the size is doubled.

1DIce: Yesterday before i answerd your questions @LorenzE told me that you guys would prefer QLinkedList for the workers, is this still true? Please let me know.

LorenzE: @1DIce: yes please use the QLinkedList

PR #318, opened 2017-10-20
Added missing binary files to .gitignore

er06645810:
There were some binary files missing from the .gitignore files so I went ahead and added them.
Just out of curiosity, is there some reason that we don't just blot out the whole binary folder with `/bin/*`?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,15 +77,17 @@
 /bin/ex_roi_clustered_inverse_pwl_rap_musicd
 /bin/ex_st_clustered_inverse_pwl_rap_music
 /bin/ex_st_clustered_inverse_pwl_rap_musicd
-
 /bin/ex_interpolation
-/bin/test_fiff_cov
+/bin/ex_interpolationd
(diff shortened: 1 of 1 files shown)


LorenzE:
@er06645810 thx for this. Regarding your question: There are some sub folders in bin/ in which we need to store resources and example data in.

er06645810:
Ah, that makes sense. In that case, you might be able to get away with an exclusion pattern such as
```
/bin/*               # Ignore everything in the bin folder
!resources/*    # Except files in the resources sub-folder 
```
That would prevent needing to update .gitignore each time a new target is added and would be more succinct. The exclusion rule can be kind of tricky to get working exactly as you want though. I often refer back to 
[How do gitignore exclusion rules actually work?](https://stackoverflow.com/questions/3001888/how-do-gitignore-exclusion-rules-actually-work)

LorenzE:
Thx for the hint. I created a new PR for this.

PR #319, opened 2017-10-20
[WIP] Don't display console on releaes builds

er06645810:
The console is extremely helpful for debugging during development, but we've decided to hide it on release builds of MNE Scan. If you find this change helpful, you're welcome to adopt it into the official repository. If not, you can just close this pull request.


LorenzE:
Let's keep the command line in release mode for now.

PR #321, opened 2017-10-24
Clean up resource handling

LorenzE:
- Create resource folder in mne-cpp folder
- Copy resources via .pro file to bin/resource folder as needed by each sub project
- Ignore everything in the bin Folder
- Refactor bci plugin


LorenzE:
@er06645810 This should also solve the gitignore issue. You would just need to copy your resource files via your plugin's .pro file to the corresponding /bin/resources/mne_scan/plugins/... Folder. Please let me know if you have questions!

er06645810:
This looks like a great way to handle it. Two thumbs up.

PR #324, opened 2017-12-07
[LIB-138] Fix transparency in Disp3D & [LIB-189] make frame graph work with OpenGL < 3.3

1DIce:
I extended the frame graph to render transparent and opaque objects correctly.
All render states used for transparency are now defined in the frame graph and were deleted form the materials.
The PerVertexPhongAlphaMaterial also has an option to sort the objects using it before rendering.

### DIFF
--- a/libraries/disp3D/engine/model/3dhelpers/geometrymultiplier.cpp
+++ b/libraries/disp3D/engine/model/3dhelpers/geometrymultiplier.cpp
@@ -117,7 +117,16 @@ void GeometryMultiplier::setTransforms(const QVector<QMatrix4x4> &tInstanceTansf
         return;
     }
 
-    m_pTransformBuffer->setData(buildTransformBuffer(tInstanceTansform));
+    //Update buffer content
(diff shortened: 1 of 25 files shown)


libraries/disp3D/engine/model/3dhelpers/geometrymultiplier.cpp
```diff
@@ -117,7 +117,16 @@ void GeometryMultiplier::setTransforms(const QVector<QMatrix4x4> &tInstanceTansf
...
+    const QByteArray &transformBufferData = buildTransformBuffer(tInstanceTansform);
+    if(transformBufferData.size() != m_pTransformBuffer->data().size())
+    {
+        m_pTransformBuffer->setData(transformBufferData);
+    }
+    else
+    {
+        m_pTransformBuffer->updateData(0, transformBufferData);
+    }
     m_pTransformAttribute->setBuffer(m_pTransformBuffer);
```
LorenzE: Do we need to do this here again? see line 182 of same file.

PR #328, opened 2018-03-16
Fix minimal version isssues

LorenzE:
- Encountered and fixed some issues when building the minimal version of MNE-CPP on a Neuromag acquisition system (Read Head Enterprise 5.1). 
- Added description back into the write_eigen_matrix function
- Renamed some build flags


chdinh:
We Actual need also adapt the travis build to Qt 5.2.1

mne-cpp.pro
```diff
@@ -37,10 +37,10 @@ include(mne-cpp.pri)
 
 TEMPLATE = subdirs
 
-#At least version 5.6.0
-!minQtVersion(5, 6, 0) {
+#At least version 5.2.0
```
chdinh: Should be 5.2.1, since beineri provides only the 5.2.1 patched version

mne-cpp.pro
```diff
@@ -37,10 +37,10 @@ include(mne-cpp.pri)
...
+!minQtVersion(5, 2, 0) {
     message("Cannot build MNE-CPP with Qt version $${QT_VERSION}.")
-    error("Use at least Qt 5.6.0.")
+    error("Use at least Qt 5.2.0. Please note that you may only be able to build the minimal MNE-CPP version.")
```
chdinh: Qt 5.2.1...

LorenzE:
Qt 5.3.0 did not build on the Neuromag system at Martinos. I tried it last week. However, Qt 5.2.0 did.

PR #350, opened 2018-07-27
Move further MNE Scan features to library layer

LorenzE:
- Rename QuickControlWidget to QuickControlView
- Outsource individual widgets in QuickControlView and make them accessible to other applications as well
- QuickControlView is now modular and can be easily adjusted to fit more individual needs
- Add function for resource copying in mne-cpp.pri
- Change lights in Disp3D View to generate an improved/realsitic visualization
- Add RtConnectivity worker in realtime library and make use of it in the Neuronal Connectivity MNE Scan plugin


applications/mne_scan/libs/scDisp/realtimeconnectivityestimatewidget.cpp
```diff
@@ -103,44 +91,48 @@ RealTimeConnectivityEstimateWidget::RealTimeConnectivityEstimateWidget(QSharedPo
...
 //*************************************************************************************************************
 
 RealTimeConnectivityEstimateWidget::~RealTimeConnectivityEstimateWidget()
 {
-    //
     // Store Settings
-    //
-    if(!m_pRTCE->getName().isEmpty())
-    {
+    if(!m_pRTCE->getName().isEmpty()) {
```
chdinh: what happens here?

PR #352, opened 2018-09-11
[ENH] Plugin Creator CLI Application

moppymopperson:
# MNE Plugin Creator

MNE Plugin Creator is a tool to automate the creation of new MNE-CPP plugins. It consists of a simple CLI that asks the user for information and then generates a project to help them get started. It also automatically creates a testframe for the new plugin to encourage proper unit testing practices.

## Motivation

Learning the Qt ecosystem is challenging and can be frustrating for new users who just want to implement a simple plugin. Even just copying the DummyPlugin and renaming it is non-trivial and surprisingly time consuming. Providing a CLI plugin creation tool the likes of Javascript's `npm` or Swift's `spm` will significantly lower MNE-CPP's barrier to entry for new users and increase its adoption.

Likewise, a detailed explanation of how the plugin interface methods work has been added to the header files of the generated plugin with the goal of bringing new developers up to speed more quickly by helping them to grasp the big picture and avoid common pitfalls.

## Usage

Presently `mne_plugin_creator` is intended to be run from within the Qt Creator IDE. Set it as the active project and run it to be presented with the CLI. You will be asked for basic information about the plugin you want to create.

When complete, Qt Creator should automatically reload the MNE-CPP workspace with your plugin included. If a pop message appears asking you if you want to save, close, or reload any files, choose the reload option.

Presently only MNE Scan's `IPlugin` and `IAlgorithm` are supported. The templates used for each are based on Christoph's DummyPlugin.

## Development

The plugin creator is designed to be extensible so that it can support more types of plugins and apps as the MNE-CPP ecosystem expands.

The basic principle behind `mne_plugin_creator` is that there are a number of plain text template files stored in it's `templates/` directory. These templates all contain fields that are marked with double brackets, such as `{{author}}` and `{{header_file_name}}`. When you create a new plugin, what actually happens is that the template files are duplicated, renamed, and the information you provided to the CLI is substituted into these fields.

Modifying the details of the generated projects is as simple as modifying the template files.

To extend `mne_plugin_creator`, you will need to start by writing a new parser class to collect information from the user. Subclass `IInputParser` and it's abstract methods. It has a number of utility function for performing common tasks like displaying options and validating input. `MNEScanInputParser` can serve as a reference. Once you've created your parser, add it to `AppInputParser`'s `parseUserInput` method.

Next you need to write an `IPluginCreator` subclass and implement its abstract methods to specify where the templates are and where to copy them to, as well as how Qt's .pro files need to be adjusted. Optionally, you can override the default implementation of the testframe creation method to provide a more detailed test frame boilerplate. `MNEScanPluginCreator` may be a valuable reference.

## Future Work

1.  Setting up unit tests for MNE Scan plugins is very difficult. We should provide a template that help users get access to their plugins inside a test frame. The present template just provides access to basic Qt classes, which isn't actually very useful at all.
    * I actually took a crack at this, but gave up after about an hour. @LorenzE might be the man for the job.

2.  There is no support for MNE Analyze plugins yet. `IInputParser` and `IPluginCreator` classes should be written for MNE Analyze.

3.  We should update `mne_plugin_creator`'s .pro file to put the build output in the MNE binary folder. The challenge is ensuring that the correct relative path to the source code folders are provided to the plugin creator classes regardless of which directory the tool is run from. Presently we assume it will be run from the default build directory and have hard coded relative paths from that directory.
    * See `MNEScanPluginCreator`'s constructor.


LorenzE:
Thx for this @moppymopperson I will take a closer look asap

moppymopperson:
@LorenzE I tested this thoroughly on Linux. I haven't run it on Windows yet. Let me know once you've tried it out.

LorenzE:
@moppymopperson Unfortunately, I did not have time yet. I will try to review it this week.

moppymopperson:
@LorenzE No rush! I just wanted to make sure you hadn't found any problems.

LorenzE:
Hi @moppymopperson . I did not forget about this and finally found some time to go over it :) This tool can be indeed very useful not only for new users. Thanks for the effort! Here are some comments and suggestions:

- The plugin works great on Windows 10. 
- As you already mentioned in your description above: The mne_plugin_creator.pro file should include deployment, copy to bin folder, etc.
- We should also consider the possibility if the repository is named different than ‘mne-cpp’. For example, I always tend to add the GitHub user name to the repositorie's folder name (mne-cpp-moppymopperson).
- Not sure if we need to ask the user about the namespace. The namespace of the plugin should always be the plugin’s name in capital letters.
- We should also think about how to handle different type of incoming measurement types. So far only RealTimeMultiSampleArray is supported.
- Since we added FFTW support not too long ago, this is still missing in the plugin template.pro file.
- We should add the resource handling to the plugin template.pro file as well. 
- Not sure if we should have default authors in the plugin’s template.pro file.
- The copyright section should include all authors names. 
- When creating an ISensor plugin it also creates a RealTimeMultiSampleArray input stream. We should maybe consider including a producer class instead.
- The overall code base looks good to me. Before merge some coding conventions need to be fixed. I can make another run and do detailed comments once all other issues have been addressed.

moppymopperson:
Thanks for the feedback. I had almost forgotten about this myself!

I'll spend some time trying to resolve the issues with hardcoded paths. Are there any files you can suggest as a good reference for how to setup deployment, copy to bin, etc.? 

If possible I want to reuse the mechanism that is already in place for resolving paths. I'll try to look at how icon resources are handled and try to emulate that.

LorenzE:
> Are there any files you can suggest as a good reference for how to setup deployment, copy to bin, etc.?

You can checkout the mne_show_fiff.pro [here](https://github.com/mne-tools/mne-cpp/blob/master/applications/mne_show_fiff/mne_show_fiff.pro). I think you should be able to more or less copy paste, change the name of your application and remove the specified resource files.

> If possible I want to reuse the mechanism that is already in place for resolving paths. I'll try to look at how icon resources are handled and try to emulate that.

Great! If you assume that the mne_plugin_creator executable always resides in the mne-cpp/bin folder how about using `QCoreApplication::applicationDirPath()`? This would also solve the problem of renamed repository folders. E.g.: 
```cpp
m_templatesPath(QCoreApplication::applicationDirPath() + "../applications/mne_plugin_creator/templates/mne_scan/")
```

applications/mne_plugin_creator/mnescanplugincreator.cpp
```diff
@@ -38,12 +38,10 @@
...
 //=============================================================================================================
 MNEScanPluginCreator::MNEScanPluginCreator()
-    : IPluginCreator()
-    , m_templatesPath("../../../mne-cpp/applications/mne_plugin_creator/templates/mne_scan/")
-    , m_pluginsPath("../../../mne-cpp/applications/mne_scan/plugins/")
-    , m_profilePath(m_pluginsPath + "plugins.pro")
-{
-}
+    : IPluginCreator(),
+      m_templatesPath(QCoreApplication::applicationDirPath() + "resources/mne_plugin_creator/templates/mne_scan/"),
```
LorenzE: QCoreApplication needs to be #included

juangpc:
Hello, 
What's the state of this pr? It has been a while since the last conversation.

moppymopperson:
I haven't worked on this PR in a couple of years. As far as I remember it was in pretty good shape and didn't need much more work. You're welcome to carry it across the finish line if you'd like, but I don't have enough bandwidth to do it myself.

PR #353, opened 2018-09-16
Improvements to averaging, connectivity and RTC-MNE plugin

LorenzE:
- Update some fiff constants
- Make Network 3D visualization more efficient by not sending data to GPU while the item is deactivated
- Make Average plugin dispatch EvokedSet as soon as a new trial was included in the computation
- Add new viewer for minimum norm settings
- Add trigger type and method selection to MNE plugin (make use of minimum norm settings viewer)
- Add trigger type selection to Connectivity plugin (when directly connected to Averaging plugin)
- Add gradiometer picking to Connectivity plugin
- Make trial rejection work on EOG channel
- Loop mode now always loops the last received data


applications/mne_scan/plugins/mne/mne.cpp
```diff
@@ -555,13 +604,19 @@ void MNE::run()
             {
                 MatrixXd rawSegment = m_pMatrixDataBuffer->pop();
 
-                float tmin = 1.0f / m_pFiffInfo->sfreq;
+                MatrixXd data(m_invOp.noise_cov->names.size(), rawSegment.cols());
```
1DIce: A small comment what the happens would be nice here.

libraries/realtime/rtProcessing/rtconnectivity.cpp
```diff
@@ -143,3 +143,12 @@ void RtConnectivity::handleResults(const Network& connectivityResult)
 {
     emit newConnectivityResultAvailable(connectivityResult);
 }
+
+
+//*************************************************************************************************************
+
+void RtConnectivity::reset()
+{
+    int returnCode;
```
1DIce: returnCode is not used.

applications/mne_scan/plugins/neuronalconnectivity/neuronalconnectivity.cpp
```diff
@@ -314,14 +316,17 @@ void NeuronalConnectivity::updateRTMSA(SCMEASLIB::Measurement::SPtr pMeasurement
             bool bPick = false;
             qint32 unit;
             int counter = 0;
-            QString sChType = "mag";
+            QString sChType = "grad";
```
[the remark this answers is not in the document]
LorenzE: that is fine for now. we need to make this parameter GUI changeable in the future.

PR #361, opened 2018-12-05
MNE Scan and MNE Browse improvements

LorenzE:
- Use OpenGL rendering in ChannelDataViewer, AverageLayoutViewer, ButterflyViewer
- Fix multiple problems in averaging GUI 
- Use signal/slot to communicate with thread when performing real-time averaging
- Start to create screenshot functionality to Disp3D via QRenderCapturer


libraries/disp/viewers/averageselectionview.cpp
```diff
@@ -98,114 +147,70 @@ void AverageSelectionView::init()
...
+        pButton->setObjectName(itr.key());
+        pButton->setStyleSheet(QString("background-color: rgb(%1, %2, %3);").arg(color.red()).arg(color.green()).arg(color.blue()));
         topLayout->addWidget(pButton, count, 1);
-        connect(pButton, &QPushButton::clicked,
+        connect(pButton.data(), &QPushButton::clicked,
                 this, &AverageSelectionView::onAveragesChanged);
-        m_qMapButtonAverageType.insert(pButton, i.value().second.first.toDouble());
 
-        ++count;
+        count++;
```
chdinh: ++count is one cpu step shorter, i.e., faster - pre-increment instead of post-increment
```suggestion
        ++count;
```

PR #375, opened 2019-05-16
Add missing RPATH

### LINKED ISSUE #297: Qt version
The [MNE-CPP Documentation](http://doc.mne-cpp.org/) states:
> ### Dependencies
> &gt;= Qt 5.7.1

A recent binary build seems to have been built against Qt 5.8:
```
$ ./mne_analyze
/neurospin/local/mne-cpp/bin/mne_analyze: /usr/lib/x86_64-linux-gnu/libQt5PrintSupport.so.5: version `Qt_5' not found (required by /neurospin/local/mne-cpp/bin/mne_analyze)
/neurospin/local/mne-cpp/bin/mne_analyze: /usr/lib/x86_64-linux-gnu/libQt5Core.so.5: version `Qt_5.8' not found (required by /neurospin/local/mne-cpp/bin/mne_analyze)
...
```

Why not build MNE-CPP against Qt 5.7.1? Does MNE-CPP actually depend on Qt 5.8?

More importantly, even the most recent version of Ubuntu LTS (16.04) ships with Qt 5.5.1. The previous version (14.04) is still maintained and ships with Qt 5.2.1. **Wouldn't it be possible to depend on one of these versions? Is there really a hard dependency on Qt 5.7.1 or 5.8?**

* Right now, deploying MNE-CPP on Ubuntu using the [Linux Binaries](https://www.mne-cpp.org/index.php/download/) is complicated: you need to build and deploy Qt as well and set LD_LIBRARY_PATH - probably for MNE-CPP only.
* MNE-CPP won't be easy to package MNE-CPP in NeuroDebian either, if it has to be bundled with its own version of Qt.

LorenzE:
Add QMAKE_RPATHDIR to MNE Scan plugins and MNE Analyze extensions as well as testframes. This will remove the necessity to set LD_LIBRARY_PATH to point to the mne-cpp lib folder on Linux. Avoiding LD_LIBRARY_PATH will make sure that no other application are linking against mne-cpp shipped qt libraries. The travis linux test script was changed to not use LD_LIBRARY_PATH. This PR fixes #297.

### DIFF
--- a/tools/travis/linux_tests/after_success.sh
+++ b/tools/travis/linux_tests/after_success.sh
@@ -6,7 +6,6 @@ set -e
 git clone https://github.com/mne-tools/mne-cpp-test-data.git mne-cpp-test-data
 
 # Set Environment variable
-export LD_LIBRARY_PATH=$(pwd)/lib:$LD_LIBRARY_PATH
 MNECPP_ROOT=$(pwd)
 
 # Tests to run - TODO: find required tests automatically with grep
--- a/tools/travis/linux_tests/after_success.sh
+++ b/tools/travis/linux_tests/after_success.sh
@@ -6,6 +6,7 @@ set -e
 git clone https://github.com/mne-tools/mne-cpp-test-data.git mne-cpp-test-data
 
 # Set Environment variable
+export LD_LIBRARY_PATH=$(pwd)/lib:$LD_LIBRARY_PATH
 MNECPP_ROOT=$(pwd)
 
 # Tests to run - TODO: find required tests automatically with grep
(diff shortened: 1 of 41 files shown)


tools/travis/linux_tests/after_success.sh
```diff
@@ -6,6 +6,7 @@ set -e
 git clone https://github.com/mne-tools/mne-cpp-test-data.git mne-cpp-test-data
 
 # Set Environment variable
+export LD_LIBRARY_PATH=$(pwd)/lib:$LD_LIBRARY_PATH
```
LorenzE: Why would we still need this? All tests are now using RPATH.

PR #385, opened 2019-06-21
solve test data paths

juangpc:
There's a (known(?)) problem with the app test_forward_solution.exe

Before, the app crashes when computing forward solution due to some path strings not well concatenated. This commit solves partially the problem, but the app is still crashing. 

As far as I understand, the forward solution is now computed (correctly?) and the app crash is now inside a call to mne_sourcespace.cpp while comparing the solution.

Feel free to delay the pull until I can figure this out.

Thanks

### DIFF
--- a/testframes/test_forward_solution/test_forward_solution.cpp
+++ b/testframes/test_forward_solution/test_forward_solution.cpp
@@ -130,7 +130,7 @@ void TestForwardSolution::computeForward()
     settings.transname.clear();
     settings.bemname = QDir::currentPath()+QCoreApplication::applicationDirPath() + "/MNE-sample-data/subjects/sample/bem/sample-5120-5120-5120-bem.fif";
     settings.mindist = 5.0f/1000.0f;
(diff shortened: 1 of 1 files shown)


LorenzE:
Yes. There seems to be a problem. The forward test is currently not included in our CI, since the forward modeling code needs to be debugged and refactored first.

testframes/test_forward_solution/test_forward_solution.cpp
```diff
@@ -130,7 +130,7 @@ void TestForwardSolution::computeForward()
     settings.transname.clear();
     settings.bemname = QDir::currentPath()+QCoreApplication::applicationDirPath() + "/MNE-sample-data/subjects/sample/bem/sample-5120-5120-5120-bem.fif";
```
LorenzE: `QCoreApplication::applicationDirPath() ` can be removed. Just use `QDir::currentPath()`

PR #388, opened 2019-07-28
Library and MNE Scan improvements

LorenzE:
Connectivity library:

- Fix some bugs when computing Coherency.
- Improved network visualization with edge diameter reflecting connection strength.
- Add example to test connectivity library performance.
- Allow larger than signal size FFT sizes (zero padding) and make according changes to the connectivity class
- Allow the selection of frequency bins to choose when computing spectral connectivity measures 
- Add storage mode in the connectivity library. If activated intermediate data is stored so that future computations are faster.
- Add ROI selection to ex_connectivity.

Disp3D library:

- Improved colormap usage.
- Added new colormaps to Disp3D: Vridis, Bone, Cool, RedBlue.
- Use rgba instead of rgb color vectors in disp3d. 
- Add setTransform and applyTransform functions to 3D tree items.
- Reduce shininess in phongalphamaterial.
- Allow transparent vertex colors.
- Add support for setting vertices with cortical activation lower than a threshold to be transparent.
- Improve sample wise averaging, looping and signal emitting in the sensor and source data worker.
- Increase light intensity of dips3d scene.

Disp library:

- Add a new GUI widget to select the row and trial number to, e.g., visualize corresponding time courses.

Fiff library:

- Add applyTransform function to FiffDigpointSet class.

Freesurfer library:

- Add function to only pick wanted labels from annotation data.

Inverse library:

- Add flag to specify to choose normal picking when calculating the inverse solution for one single data matrix with the MinimumNorm class.

MNE library:

- Introduce QMap to store channel types and thresholds to reject trials in MNEEpochDataList.
- Add convenient function to get row indices of sources in stc matrix based on picked labels in the MNESourceEstimate class.

Fwd library:

- Add convenient function to receive source positions in 3D space by label from a forward solution.

MNE Scan:

- Get rid of thread wait routines which did not account for the fact that functions from classes, which inherit QThread, are running in the caller’s thread. Only the run method of a class inheriting from QThread actually runs in a new thread.
- Change MNE plugin to pick normal when reducing the dipole moments. This way we keep the radial components.


libraries/fiff/fiff_dig_point_set.cpp
```diff
@@ -299,3 +299,24 @@ FiffDigPointSet &FiffDigPointSet::operator<<(const FiffDigPoint *dig)
     this->m_qListDigPoint.append(*dig);
     return *this;
 }
+
+//*************************************************************************************************************
+
+void FiffDigPointSet::applyTransform(const FiffCoordTrans& coordTrans, bool bApplyInverse)
```
chdinh: Which kind of transform is this? Didn't we had this before?

libraries/inverse/minimumNorm/minimumnorm.cpp
```diff
@@ -192,7 +192,7 @@ MNESourceEstimate MinimumNorm::calculateInverse(const FiffEvoked &p_fiffEvoked,
 
 //*************************************************************************************************************
 
-MNESourceEstimate MinimumNorm::calculateInverse(const MatrixXd &data, float tmin, float tstep) const
+MNESourceEstimate MinimumNorm::calculateInverse(const MatrixXd &data, float tmin, float tstep, bool pick_normal) const
```
chdinh: Is the picknormal checked with the python version?

libraries/disp/plots/helpers/colormap.h
```diff
@@ -168,14 +427,45 @@ class DISPSHARED_EXPORT ColorMap
 
     //=========================================================================================================
     /**
-    * Returns a RedBlue RGB to a given double value [-1,1]
+    * Returns a Bone RGB to a given double value [-1,1]
```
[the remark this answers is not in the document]
LorenzE: Will fix this

libraries/fiff/fiff_dig_point_set.cpp
LorenzE: No, this functionality was not present before.

libraries/inverse/minimumNorm/minimumnorm.cpp
LorenzE: No testing vs mne-python was performed

libraries/disp/plots/helpers/colormap.h
```diff
@@ -73,6 +76,262 @@
 namespace DISPLIB
 {
 
+const Eigen::MatrixX3i m_matViridrisData = (Eigen::MatrixX3i(256,3) << 68, 1, 84,
```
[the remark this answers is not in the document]
LorenzE: Viridis's RGB functions are highly non linear. The authors from matplotlib published RGB values, which can be used as a lookup table. See this comment for a discussion on this: https://github.com/JuliaPlots/Plots.jl/issues/185#issuecomment-209518858

libraries/disp/plots/helpers/colormap.h
```diff
@@ -447,6 +791,41 @@ inline QRgb ColorMap::valueToRedBlue(double v)
...
+
+//*************************************************************************************************************
+
+inline QRgb ColorMap::valueToViridis(double v)
+{
+    if(v*255 >= m_matViridrisData.rows()) {
+        return QRgb();
+    }
+
+    QRgb p_qRgb = qRgb(m_matViridrisData(v*255,0), m_matViridrisData(v*255,1), m_matViridrisData(v*255,2));
```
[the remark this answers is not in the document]
LorenzE: Will fix this

LorenzE:
Agreed, the scope of this PR is too wide. Next ones will be more granular again :)

PR #393, opened 2019-08-27
[ENH] mne_anonymize

juangpc:
An application to anonymize fiff files. To Do:
- [x] Finish unit test test_mne_anonymize. Come up with a testing function
- [x] Move applicaton logic from FiffAnonymizer to SettingsController class
- [x] Fix multiple file inputs
- [x] Figure out what to do with birthday substractions. mne-python's anonymizer is handling this differently.
- [x] Delete FiffDir functionality and do not write this at the end of the file
- [x] Create documentation page. For preview click [here](http://lorenze.github.io/mne-cpp/pages/learn/mneanonymize.html)


applications/mne_fiff_anonymizer/mne_fiff_anonymizer.pro
```diff
@@ -0,0 +1,78 @@
...
+#
+# @brief    This project file generates the makefile to build the mne_fiff_anonymizer example.
+#
+#--------------------------------------------------------------------------------------------------------------
+
+include(../../mne-cpp.pri)
+
+TEMPLATE = app
+
+QT += widgets 3dextras
```
LorenzE: Since you included your application in the mininmalVersion of mne-cpp here, `3dextras` from the Qt3D module is not available. Also  we do not `3dextras` because do not link against the disp3d library. So you can safely remove `3extras` here. Also the minimalVersion should only include non GUI related tools and libraries. Consider removing `widgets` as well or move the application to the non minimalVersion in applications.pro file.

juangpc: I didn't know I had included it in the minimalVersion, however if it is there I think it might come in handy to someone (hope). It is only command-line. @imsorryk was going to remove 3dextras and widgets. Thanks for the comment. (I'll ask you sometime later how did I include the app in the minimal version).

applications/mne_fiff_anonymizer/main.cpp
```diff
@@ -0,0 +1,83 @@
...
+// Eigen
+//=============================================================================================================
+
+
+//*************************************************************************************************************
+//=============================================================================================================
+// QT INCLUDES
+//=============================================================================================================
+
+#include <QApplication>
```
LorenzE: This is breaking the minimalVersion CI build on travis. Using `#include <QCoreApplication>` instead should fix this. This happens due to the fact that `QApplication` is part of the `widgets` module, which now has been removed from the `mne_fiff_anonymizer.pro` file in your last commit.

applications/mne_fiff_anonymizer/mne_fiff_anonymizer.pro
LorenzE: Alright. Let's keep it in the minimalVersion for now. If you want to create a small GUI mode in the future we can always bump it up to the full version.

applications/mne_fiff_anonymizer/main.cpp
juangpc: I actually just noticed that. Talking about functional connectivity...

applications/mne_fiff_anonymizer/commandlineuser.cpp
```diff
@@ -0,0 +1,154 @@
...
+    parser.addHelpOption();
+    parser.addVersionOption();
+
+    parser.addPositionalArgument(
+                "Input File",QCoreApplication::translate("main","File to anonymize"),"File Path");
+
+
+    QCommandLineOption outputOption("fileOut", "The output file <out>.", "out", QCoreApplication::applicationDirPath() + "/MNE-sample-data/MEG/sample/testing_output.fif");
+
+    parser.addOption(inputOption);
```
LorenzE: `inoutOption` is undeclared

juangpc: :trollface:

applications/mne_fiff_anonymizer/main.cpp
```diff
@@ -0,0 +1,89 @@
...
+*
+* @param [in] argc (argument count) is an integer that indicates how many arguments were entered on the command line when the program was started.
+* @param [in] argv (argument vector) is an array of pointers to arrays of character objects. The array objects are null-terminated strings, representing the arguments that were entered on the command line when the program was started.
+* @return the value that was set to exit() (which is 0 if exit() is called via quit()).
+*/
+int main(int argc, char *argv[])
+{
+    QCoreApplication qtApp(argc, argv);
+    MNEFIFFANONYMIZER::FiffAnonymizerApp app;
+    MNEFIFFANONYMIZER::CommandLineUser user{&app,&qtApp};
```
LorenzE: `{...}` style inits are not allowed in C++98. We are dependent on C++98 standard in the minimalVersion since we need to be able to compile the minimalVersion on the old Neuromag acq systems.

applications/mne_fiff_anonymizer/fiffanonymizerapp.h
```diff
@@ -0,0 +1,202 @@
...
+    QString m_sDfltProjectPersons;
+    QString m_sDfltProjectComment;
+
+    QString m_sFileNameIn;
+    QString m_sFfileNameOut;
+
+    QFile* m_fFileIn;
+    QFile* m_fFileOut;
+
+    QDebug *m_printInSameLineHelper;
```
LorenzE: Try to avoid raw pointers as much as possible. If you still need to use them make sure you handle the garbage collection, i.e. delete them in the class' destructor. I would recommend using a guarded pointer `QPointer<QDebug>`, see [docu](https://doc.qt.io/qt-5/qpointer.html#details). Another benefit when using `QPointer` is that initializes to `Q_NULLPTR` by default and when delete was called. This way checking the pointer integrity via `if(pointer){...}` will never crash.

Why is a pointer needed for this member? Why don't you use a simple `qDebug() << ..` in `printIfVerbose()`?

applications/mne_fiff_anonymizer/fiffanonymizerapp.cpp
```diff
@@ -0,0 +1,631 @@
...
+      m_sDfltSubjectComment(m_sDfltString),
+      m_sDfltSubjectHisId(m_sDfltString),
+      m_iDfltProjectId(0),
+      m_sDfltProjectName(m_sDfltString),
+      m_sDfltProjectAim(m_sDfltString),
+      m_sDfltProjectPersons(m_sDfltString),
+      m_sDfltProjectComment(m_sDfltString),
+      m_sFileNameIn(""),
+      m_sFfileNameOut(""),
+      m_printInSameLineHelper(nullptr),
```
LorenzE: `nullptr `was introduced in C++11, which is not supported by the minimalVersion. Use `Q_NULLPTR` instead to init raw pointers. If you decide to use a guarded pointer (see comment below) this line of code is obsolete.

applications/mne_fiff_anonymizer/fiffanonymizerapp.cpp
```diff
@@ -0,0 +1,631 @@
...
+        qDebug() << "You have requested to delete the input file: " + m_fFileIn->fileName();
+        if(m_bDeleteInputFileConfirmation)
+        {
+            QTextStream consoleOut(stdout);
+            QTextStream consoleIn(stdin);
+            QString confirmation;
+            qDebug() << "You can avoid this confirmation by using the delete_confirmation option.";
+            consoleOut << "Are you sure you want to delete this file? [Y/n] ";
+            consoleIn >> confirmation;
+            if(confirmation == 'Y') {
```
LorenzE: use `"Y"` instead of `'Y'`

applications/mne_fiff_anonymizer/fiffanonymizerapp.h
juangpc: You spotted a problem. Solved.
My intention is to be able to use qDebug and print to the same line. Say you want to print:
"Reading file..."
Then perform the reading and depending on the result have the user see:
"Reading file... OK" or  "Reading file... ERROR"
I store the "previously used QDebug object" in m_printInSameLineHelper, maybe I should've called it m_qdebugPreviousUseHelper. or something of the sort.

applications/mne_fiff_anonymizer/fiffanonymizerapp.cpp
juangpc: it is no longer a pointer, so no it is initialized with a qDebug() call to generate a QDebug obj.
Thanks.

applications/mne_fiff_anonymizer/fiffanonymizerapp.cpp
juangpc: Thanks.

applications/mne_fiff_anonymizer/fiffanonymizerapp.h
LorenzE: Great read on logging in Qt is also available [here](https://www.kdab.com/wp-content/uploads/stories/slides/Day2/KaiKoehne_Qt%20Logging%20Framework%2016_9_0.pdf)

juangpc: Thank you for the read. That is great.
I'm a bit confused. I thought c++ people discourage the use of macros, and promote the use of enum class and const or constexpr for simple variables. I'm surprised how Qt actively uses them. Sometimes in mne-cpp you can see things like "return (expr);" which remind me of the problems with macros.

applications/mne_anonymize/fiffanonymizer.cpp
```diff
@@ -0,0 +1,724 @@
...
+    m_BDfltMAC[7] = 0x00;
+
+    m_pBlockTypeList = QSharedPointer<QStack<int32_t> >(new QStack<int32_t>);
+    m_pOutDir = QSharedPointer<QVector<FiffDirEntry> >(new QVector<FiffDirEntry>);
+}
+
+
+//*************************************************************************************************************
+
+FiffAnonymizer::FiffAnonymizer(const FiffAnonymizer& obj)
```
LorenzE: This is just a first version of the copy constructor. We need get methods next to all set methods to retreive each to be copied parameter of obj here.

juangpc: Thanks for your last pushes. Yes. I'll try to do this today.

juangpc:
@LorenzE could we set a flag or something of the sort, to be able to push modifications without triggering a testing procedure?
I just would like to keep pushing stuff every now and then s that @imsorryk and yourself can be updated.
Other ideas> - drop the pr until it is ready.
                      - branch to mne_fiff_anonymizerDEV. 
what do you think?

LorenzE:
> @LorenzE could we set a flag or something of the sort, to be able to push modifications without triggering a testing procedure?
> I just would like to keep pushing stuff every now and then s that @imsorryk and yourself can be updated.
> Other ideas> - drop the pr until it is ready.
> - branch to mne_fiff_anonymizerDEV.
> what do you think?
> [edit] or just keep working... 🤕

@jasmainak just told me that you can use `[skip ci]` in your commit line as described [here for TravisCis](https://docs.travis-ci.com/user/customizing-the-build/#skipping-a-build) and [here for Appveyor](https://www.appveyor.com/docs/how-to/filtering-commits/).

LorenzE:
1) I think we should keep the same naming of command line flags as in the mne-c version, see screenshot from the mne-c manual below. 
2) How about  modifying the behavior to match the mne-c version, meaning that if no output file is specified, the application will overwrite the file.
3) Should we add a flag to handle the FIFF_SUBJ_HIS_ID tag as in the mne-c version.
image

juangpc:
**I think we should keep the same naming of command line flags as in the mne-c version.**
   -> Yes. This was my intention all along. 

**How about modifying the behavior to match the mne-c version, meaning that if no output file is specified, the application will overwrite the file.**
   -> I'm more neutral on this one. As a user of the app, I think it will be easy to "understand" the app either way. Besides, I think usual gnu applications have a default output filename and only if you specify the option (typically, -o or -out) you can specify the output name. We had this conversation at the beginning and we decided that to modify the input file would be too dangerous as users could inadvertently modify their data. And I think that this mne-behavior is more related to the fact that mne's tool does the anonymization **in-place**. Once we adopted the idea of doing it modifying the length of each field, thus altering the file, thus creating a new file... It is, in my opinion, more natural, to create a different output file.
Right now: 
- if you don't specify -out option. output file name will be equal to input filename pluse "_anonymized".
-if you do specify the option, that will be used. 
-if you want the same name of the file, specify output filename as input filename.
-if you want the input file deleted --> specify so. You will be prompted with a confirmation.
-if you want the input file deleted without confirmation --> specify so.

But, if you think we should do this, lets do it right now. Easy to modify it. No look back.

**Should we add a flag to handle the FIFF_SUBJ_HIS_ID tag as in the mne-c version.** 
Right now, this is modified only if the "-brute" flag is used. In this case, the subject id is an int32 modified to be 0.  It is really easy to add an input option "his" and if it is set, use the input and insert it into the field. Really easy right now. I think we should do this.

LorenzE:
Let's keep creating a new file by default. I agree in place anonymization can lead to somewhat "dangerous" situations. 

Great, let's add the `--his` flag then.

LorenzE:
@juangpc Do you want me to add the raw structure for a test under /testframes/?

juangpc:
@LorenzE That would be great.

LorenzE:
I added a first version of the test. It does not do much yet except for calling the new cmd line tool via QProcess.

LorenzE:
The FiffDirPointer is currently set/written to the new file. This means while reading the new/anonymozed file back into mne-cpp via FiffRawData it will try to locate the FiffDir in the file. This leads to a crash with new/anonymized files [here](https://github.com/mne-tools/mne-cpp/blob/a53138393a5a28e834556b2823c8fdc826044a67/libraries/fiff/fiff_stream.cpp#L331). @juangpc  Can you please have a look on how you handle the fiff directory writing? Do you update the position of the FiffDir in the file?

LorenzE:
@juangpc had you have time to work on this?

juangpc:
Hi, I am trying to get this out the door right now....

LorenzE:
@juangpc and @imsorryk  Want to take a final look at this? PR is ready from my side. I am currently writing up some examples for our documentation, which I will push in a subsequent PR.

LorenzE:
I added a first draft of the documentation. You can check it out [here](http://lorenze.github.io/mne-cpp/pages/learn/mneanonymize.html). 

Also, I just realized that we are missing the `FIFF_SUBJ_SEX` tag. @juangpc and @imsorryk: Is there a specific reason for this? If not, I will add the missing tag.

PR #397, opened 2019-09-20
[ENH] EDF-to-Fiff Application

sheinke:
Added a new application edf2fiff that is able to convert EDF files into Fiff files. edf2fiff is able to handle multiple frequencies but ignores interrupted recordings (i.e. treats them as uninterrupted). Raw value scaling can be directly set through commandline arguments.
The newly added coded is tested in a testframe edf2fiff_rwr.


juangpc:
Hi @sheinke, I've been working on a fiff anonymizing tool lately. If you don't mind, I would like to ask you just a question, out of curiosity. When you say that interrupted recordings are treated as uninterrupted, isn't there some kind of trigger signal to mark an interruption?

Great tool, by the way!

applications/mne_edf2fiff/main.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+        iSamplesRead += iNextChunkSize;
+
+        outfid->write_raw_buffer(data, cals);
+    }
+
+    outfid->finish_writing_raw();
+    qDebug() << "Writing finished !";
+
+    return 0;
```
juangpc: Instead of letting QCoreApp instance a, to drop out of scope and directly call its destructor, is there a reason why not calling the a.exec() method so that return can capture its output? I've wondered about this myself. I decided to follow the documentation, although anonymizer uses zero of the infrastructure that qt provides for event handling, wierd cases (like if the user logs off while the program is running, etc... ).

sheinke: Hi @juangpc , thanks for your feedback ! Frankly, I have forgotten why I chose not to run the exec-function for the QCoreApplication. I think I saw this in a similar scenario elsewhere in the codebase. As long as the Qt-Backend is not used, this should be fine.

Regarding your first comment: Interrupted recordings will produce some sort of "cut" in the signals, meaning that the values will have a spike on the beginning of the new recording, respectively. Whether or not the recording contains a trigger signal for marking interruptions is fairly irrelevant for the Fiff-conversion. The EDF+ standard also supports interrupted recordings, I just wanted to clarify that this feature is (so far) not supported and that all EDF files are treated as if they were perfectly continous.

juangpc: Interesting.
Thanks.

gabrielbmotta:
rebased and merged in #760

PR #400, opened 2019-10-02
Add test for fwd library

LorenzE:
Add test for the fwd library. This is still very much WIP. 

- [x] Overload == operator to quickly check if the forward solution and the included (complex) objects are equal
- [ ] Refactor some C code to match when reading the fwd solution back into mne-cpp

### DIFF
--- a/libraries/fiff/fiff_ch_info.cpp
+++ b/libraries/fiff/fiff_ch_info.cpp
@@ -95,3 +95,4 @@ FiffChInfo::~FiffChInfo()
 {
 
 }
+
--- a/libraries/fiff/fiff_ch_info.h
+++ b/libraries/fiff/fiff_ch_info.h
@@ -119,6 +119,16 @@ public:
     */
     inline static qint32 storageSize();
 
+    //=========================================================================================================
+    /**
+    * Overloaded == operator to compare an object to this instance.
+    *
+    * @param[in] object    The object whisch should be compared to.
+    *
+    * @return true if equal, false otherwise
+    */
+    friend bool operator== (const FiffChInfo &a, const FiffChInfo &b);
+
 public:
     fiff_int_t    scanNo;       /**< Scanning order number 1*/
     fiff_int_t    logNo;        /**< Logical channel # 1*/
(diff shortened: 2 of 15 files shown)


libraries/fwd/computeFwd/compute_fwd.cpp
```diff
@@ -1893,44 +1893,45 @@ ComputeFwd::~ComputeFwd()
...
+//    fiffId         mri_id  = NULL;
+//    fiffId         meas_id = NULL;
+//    FILE           *out = NULL;     /* Output filtered points here */
+
+//    FwdCoilSet*       templates = NULL;
+//    FwdEegSphereModel* eeg_model = NULL;
+//    FwdBemModel*       bem_model = NULL;
+
+//    QString qPath;
+//    QFile file;
```
chdinh: What happened here?

libraries/fwd/computeFwd/compute_fwd.cpp
```diff
@@ -2248,33 +2267,35 @@ void ComputeFwd::calculateFwd() const
...
+//            delete meg_forward;
+//        if(eeg_forward)
+//            delete eeg_forward;
+//        if(meg_forward_grad)
+//            delete meg_forward_grad;
+//        if(eeg_forward_grad)
+//            delete eeg_forward_grad;
+
+//        if (!res)
+//            qCritical("err_print_error();");//err_print_error();
```
chdinh: And why is this commented out?

libraries/fwd/fwd_bem_model.cpp
```diff
@@ -321,13 +327,37 @@ float **mne_mat_mat_mult_40 (float **m1,float **m2,int d1,int d2,int d3)
...
         }
+    }
+
+//    for (j = 0; j < d1; j++)
+//        for (k = 0; k < d3; k++) {
+//            sum = 0.0;
+//            for (p = 0; p < d2; p++)
+//                sum = sum + m1[j][p]*m2[p][k];
+//            result[j][k] = sum;
+//        }
```
chdinh: has this been tested?

LorenzE: yes. the test returns the same results for this hack.

libraries/fwd/computeFwd/compute_fwd.cpp
LorenzE: This only temporary until we have the fwd library refactored. This is only done in order to provide easy testing and access to the results in test_forward_solution. This is a workaround because the reading and writing is incompatible between our old and new (mne-c) MNEForwardSolution/MNESourceSpace classes.

libraries/fwd/computeFwd/compute_fwd.cpp
LorenzE: see comment above. We should not delete here since we want to access the data via test_forward_solution after we wrote to file. This is temporary as well.

PR #401, opened 2019-10-17
Add equal operator to MNEHemisphere and MNESourceSpace

LorenzE:
Add equal operator to MNEHemisphere and MNESourceSpace and use them in test_mne_forward_solution

### DIFF
--- a/libraries/fiff/fiff_named_matrix.h
+++ b/libraries/fiff/fiff_named_matrix.h
@@ -316,7 +316,7 @@ inline bool operator== (const FiffNamedMatrix &a, const FiffNamedMatrix &b)
             a.ncol == b.ncol &&
             a.row_names == b.row_names &&
             a.col_names == b.col_names &&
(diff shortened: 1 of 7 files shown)


libraries/fiff/fiff_named_matrix.h
```diff
@@ -316,7 +316,7 @@ inline bool operator== (const FiffNamedMatrix &a, const FiffNamedMatrix &b)
             a.ncol == b.ncol &&
             a.row_names == b.row_names &&
             a.col_names == b.col_names &&
-            a.data.isApprox(b.data, 0.0001f));
+            a.data.isApprox(b.data, 0.0001));
```
chdinh: why is the float qualifier removed?

LorenzE: because data is an eigen matrix of type double not float. Compiler was throwing a warning.

PR #414, opened 2019-12-03
Coverity changes

gabrielbmotta:
- initialized uninitialized variables in constructors
- added else/defaults to prevent unknown behavior
- removed throw() that was not caught in any of the examples
- changed rand() to qt equivalent
- other minor changes

### DIFF
--- a/applications/mne_matching_pursuit/mainwindow.cpp
+++ b/applications/mne_matching_pursuit/mainwindow.cpp
@@ -1558,8 +1558,8 @@ void MainWindow::recieve_result(qint32 current_iteration, qint32 max_iterations,
     else if(adaptive_atom_res_list.isEmpty())
     {
         FixDictAtom temp_atom = fix_dict_atom_res_list.last();
-        fix_dict_atom_res_list.last().display_text = create_display_text(&fix_dict_atom_res_list.last());
-        temp_atom.display_text =  create_display_text(&fix_dict_atom_res_list.last());
+        fix_dict_atom_res_list.last().display_text = create_display_text(fix_dict_atom_res_list.last());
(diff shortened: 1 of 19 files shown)


applications/mne_matching_pursuit/mainwindow.cpp
```diff
@@ -2039,83 +2040,83 @@ void MainWindow::calc_fix_mp(QString path, MatrixXd signal, truncation_criterion
 
 //*************************************************************************************************************
 
-QString MainWindow::create_display_text(FixDictAtom global_best_matching)
+QString MainWindow::create_display_text(FixDictAtom *global_best_matching)
```
LorenzE: why not use a const reference? `(const FixDictAtom& global_best_matching)`

applications/mne_matching_pursuit/mainwindow.h
```diff
@@ -797,7 +797,7 @@ private slots:
     *
     * @return       QString                 display text
     */
-    QString create_display_text(FixDictAtom global_best_matching);
+    QString create_display_text(FixDictAtom *global_best_matching);
```
LorenzE: why not const reference?

examples/ex_clustered_inverse_mne_raw/main.cpp
```diff
@@ -436,11 +437,10 @@ int main(int argc, char *argv[])
     // Calculate the average
     // Option 1 - Random selection
     VectorXi vecSel(50);
-    srand (time(NULL)); // initialize random seed
 
     for(qint32 i = 0; i < vecSel.size(); ++i)
     {
-        qint32 val = rand() % count;
```
LorenzE: what kind of coverity warning is rand() giving?

PR #417, opened 2019-12-05
Coverity part3

gabrielbmotta:
- variable initialization in constructors
- casting operands
- checks and warning messages for variables obtaining data

### DIFF
--- a/libraries/disp3D/engine/model/data3Dtreemodel.cpp
+++ b/libraries/disp3D/engine/model/data3Dtreemodel.cpp
@@ -419,7 +419,12 @@ SensorSetTreeItem* Data3DTreeModel::addMegSensorInfo(const QString& sSubject,
 
     if(!itemList.isEmpty() && (itemList.first()->type() == Data3DTreeModelItemTypes::SensorSetItem)) {
         pReturnItem = dynamic_cast<SensorSetTreeItem*>(itemList.first());
-        pReturnItem->addData(sensor, lChInfo, "MEG", bads, m_pModelEntity);
+        if(pReturnItem == NULL || pReturnItem == nullptr || pReturnItem == Q_NULLPTR){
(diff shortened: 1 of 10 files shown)


libraries/disp3D/engine/model/items/digitizer/digitizersettreeitem.cpp
```diff
@@ -241,6 +241,11 @@ void DigitizerSetTreeItem::addData(const FIFFLIB::FiffDigPointSet& tDigitizer, Q
     for(int i = 0; i < itemList.size(); ++i) {
         DigitizerTreeItem* item = dynamic_cast<DigitizerTreeItem*>(itemList.at(i));
 
+        if(item == NULL || item == nullptr || item == Q_NULLPTR){
```
LorenzE: How about `if(item == Q_NULLPTR){` ? Do we always need to check for all three? Q_NULLPTR should either be null or nullptr depending on whether C++11 flags are set or not.

libraries/disp3D/engine/model/data3Dtreemodel.cpp
```diff
@@ -449,7 +453,12 @@ SensorSetTreeItem* Data3DTreeModel::addEegSensorInfo(const QString& sSubject,
 
     if(!itemList.isEmpty() && (itemList.first()->type() == Data3DTreeModelItemTypes::SensorSetItem)) {
         pReturnItem = dynamic_cast<SensorSetTreeItem*>(itemList.first());
-        pReturnItem->addData(tempBem, lChInfo, "EEG", bads, m_pModelEntity);
+        if(pReturnItem == NULL || pReturnItem == nullptr || pReturnItem == Q_NULLPTR){
```
LorenzE: `if(pReturnItem == Q_NULLPTR){`

PR #431, opened 2019-12-18
Fix jom install in devbuilds.yml

LorenzE:
Forgot to remove if statement from non matrix based devbuild.yml workflow -_-


.github/workflows/devbuilds.yml
```diff
@@ -122,7 +122,6 @@ jobs:
         arch: win64_msvc2017_64
         modules: qtcharts
     - name: Install jom (Windows)
-      if: matrix.os == 'windows-2016'
```
buildqa: OK for me

PR #450, opened 2020-01-22
Make tests not run on gh-page changes

gabrielbmotta:
added 'paths-ignore' to pullrequest.yml to not run on doc/gh-pages

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -4,12 +4,14 @@ on:
   pull_request:
     branches:
     - master
+    paths-ignore:
+    - 'doc/gh-pages/**'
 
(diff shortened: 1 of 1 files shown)


LorenzE:
@gabrielbmotta Were you able to find a solution for this? I do not think this is a high priority task right now. So if this is stale please let me know and create an create an issue so we don't forget about it ;) I will then close the PR for now.

gabrielbmotta:
> @gabrielbmotta Were you able to find a solution for this? I do not think this is a high priority task right now. So if this is stale please let me know and create an create an issue so we don't forget about it ;) I will then close the PR for now.

No solution yet. I've made an issue like you suggested.

PR #456, opened 2020-01-30
Open website-links in new tabs

RDoerfel:
Open website links in new tab in browser by adding `{:target="_blank" rel="noopener"}` after every link.

Example: `[Github](https://github.com/mne-tools/mne-cpp){:target="_blank" rel="noopener"}`

Explanation:

- `target = "_blank"` opens the linked document in a new window or tab
- `rel = "noopener"` for security reasons [read here](https://dev.to/gulshansaini/what-is-rel-noopener-1214)

Note:
Develop->API/Documentation still opens in same tab. It seems like we are not able to open in new tab with this way of redirecting


LorenzE:
Thx @RDoerfel . Did you inluce the example by accident?

PR #477, opened 2020-02-10
Fix mutexing in ApplicationLogger class

### LINKED ISSUE #282: BUG: Log messages not thread-safe?

LorenzE:
I did some more digging on thread safety as proposed in #282. `qDebug()`, `qInfo()` etc. are documented as thread safe, see [here](https://doc.qt.io/qt-5/qtglobal.html#qDebug). So in my opinion a `QMutex` is only necessary in order to not mix up different messages from different threads in the console output. However, usage of a `QMutex` is discouraged [here](https://stackoverflow.com/questions/22527253/is-qdebug-thread-safe), which states: 

> NOTE: as Kuba Ober correctly noted, this should be used with caution (like any locking in general, though). For example, you can get a deadlock if QtDebug functions are called from I/O library internals while using same I/O library to output debug messages (this is possible, for example, when QtDebug message handler calls I/O while holding a lock on non-recursive mutex, then underlying I/O machinery calls some callback function, and then this callback function calls QtDebug function, which calls the same handler again).

and 

> It should be noted that this is possibly a terrible idea. Why? Because if you end up invoking the message handler while other mutexes are held, you may end up with a deadlock as there's no guarantee that the mutexes would be acquired in the order necessary to prevent deadlocks. For example, if the standard output is used anywhere else in your code, the C library may hold a mutex of its own, etc. It requires great care to insure that this code won't deadlock. It's possible, but don't get the idea that you can do this and it'll "just work".

This PR changes:
- Removed mutex for now.
- Implemented feature for date/time to be added to log message.
- Added logger to MNE Scan.
- Added default statement and used default color in switch statement.

@RDoerfel Can you please take a look at this?

Closes #282


libraries/utils/generics/applicationlogger.cpp
```diff
@@ -86,20 +87,29 @@ using namespace std;
     #define COLOR_FATAL         "\033[33;1m"        //yellow
     #define COLOR_RESET         "\033[0m"           //reset
 
-    #define LOG_WRITE(OUTPUT, COLOR, LEVEL, MSG) OUTPUT << COLOR << "" LEVEL " " << COLOR_RESET << MSG << "\n"
```
RDoerfel: @LorenzE Have you checked if there is still a white space or separation between keyword and actual massage?

LorenzE: yes there is

PR #486, opened 2020-02-13
[ENH] Add BrainFlowBoard plugin to MNE-SCAN

Andrey1994:
Signed-off-by: Andrey Parfenov <a1994ndrey@gmail.com>

Restore changes from #451 in new PR.

Also I changed sleeptimeout and now it should add ~10 values per one call
```
int iMinSamples = 10;
unsigned long lSamplingPeriod = (unsigned long)(1000000.0 / m_iSamplingRate) * iMinSamples;
```

PR #487, opened 2020-02-13
[ENH] Add BrainFlowBoard plugin to MNE-SCAN

Andrey1994:

Reopen #486  from another branch


LorenzE:
Does the BrainFlow code add scaling to the cahnnels already? 

image

Andrey1994:
for synthethic board it returns sin with amplitude 10000 as I remember, so to visualize it correctly in setting widget you need to set scale to smth around 10k

Andrey1994:
it should look like this for synthethic board https://brainflow.readthedocs.io/en/stable/notebooks/denoising.html if filters are disabled

LorenzE:
The plugin running without subsequent plugins seems to work now. 

Without filtering:

image

With filtering:

image

However, there is a problem when forwarding blocks with different sizes/lengths to subsequent plugins. Our circular buffers between the plugins only support blocks with equal sizes.

Andrey1994:
<< there is a problem when forwarding blocks with different sizes/lengths to subsequent plugins. Our circular buffers between the plugins only support blocks with equal sizes.
Can it be solved by adding intermediate circular buffer? Can it be done in another PR?

LorenzE:
> Can it be solved by adding intermediate circular buffer? Can it be done in another PR?

Yes, I think we might need an extra buffer to fill a matrix of fixed size.

doc/gh-pages/pages/development/brainflow.md
```diff
@@ -15,7 +15,11 @@ This plugin adds BrainFlow data acquisition SDK to MNE-SCAN app.
 
 ## Compilation of BrainFlow submodule
 
-* Make sure that you have brainflow git submodule
+* Make sure that you have brainflow git submodule by typing
+
+    $ git submodule init
```
Andrey1994: it looks like you missed code block symbols here "```" and I have these commands below

applications/mne_scan/plugins/brainflowboard/README.md
```diff
@@ -0,0 +1,29 @@
+# BrainFlow Plugin
```
Andrey1994: and this file looks a little outdated, in my latest commits its removed and replaced by doc page. We can keep readme as well but lets make it like page from docs

doc/gh-pages/pages/development/brainflow.md
Andrey1994: I fixed it, I hope you dont mind

LorenzE: not at all. Thx

applications/mne_scan/plugins/brainflowboard/README.md
LorenzE: ups, yes you are right. The README.md should be removed from PR.

Andrey1994: removed

Andrey1994:
I think everything is oknow, I would like to add an intermediate buffer and a few more changes in next PRs

LorenzE:
> I think everything is oknow, I would like to add an intermediate buffer and a few more changes in next PRs

Agreed 👍 Also, I added a first version of the buffer to the run method in my last commit.

https://github.com/Andrey1994/mne-cpp/blob/2ea27a4259fac644f1246b83388c97d64e89e84f/applications/mne_scan/plugins/brainflowboard/brainflowboard.cpp#L402

Andrey1994:
Cool, thanks!

Andrey1994:
Seems like during force push you overrided several important changes.
For example vert scale field in setup widget is removed in my master branch but exists in this one. 
Also this code
```
for (int i = 0; i < m_iNumberChannels; i++)
    {
        m_outputConnectors.pop_back();
    }
```
is wrong and doesnt exist in master but exists in PR branch

Andrey1994:
fixed, case:
prepare_session -> release_session -> prepare_session was broken.
Now it works

LorenzE:
> Seems like during force push you overrided several important changes.
> For example vert scale field in setup widget is removed in my master branch but exists in this one.
> Also this code
> 
> ```
> for (int i = 0; i < m_iNumberChannels; i++)
>     {
>         m_outputConnectors.pop_back();
>     }
> ```
> 
> is wrong and doesnt exist in master but exists in PR branch

oops thx for pushing the fix commit. 

Once the CI pipeline is fixed we will start merging PRs again, see miurahr/aqtinstall#91.

Andrey1994:
@LorenzE thanks for merging!

I have two more questions:
* For real boards its nice to have bandstop(notch) filter but there is no such field in filter design widget, but I see something about notch filter here https://github.com/mne-tools/mne-cpp/blob/master/libraries/disp/viewers/filterdesignview.cpp#L222. So the question is, how can I add/use notch filter? And is it possible to apply several filters?(bandstop + bandpass)
* What do you think about changing CI pipelines a little to build this plugin automatically?

LorenzE:
@Andrey1994 Notch filters are supported by our filter design algorithms. I do not recall why did not provide support for them in the FilterDesign GUI, see [here](https://github.com/mne-tools/mne-cpp/blob/17e95ab5d7aff9153ece7d26385f38a20acf2bb7/libraries/disp/viewers/filterdesignview.cpp#L447). Multiple filters are supported by the [API](https://github.com/mne-tools/mne-cpp/blob/17e95ab5d7aff9153ece7d26385f38a20acf2bb7/libraries/rtprocessing/rtfilter.cpp#L96), but are not used in MNE Scan currently.

I think changing the CI to also include bulding/distributing plugins is a good idea. Of course this can only be done for plugins, which need freely available SDKs, such as yours. Do you want to setup a PR draft for this? We could then work together on this.

PR #504, opened 2020-02-27
[Fix] Custom Logger and small refactoring

### LINKED ISSUE #500: Application Logger breaks on Win7
https://github.com/mne-tools/mne-cpp/blob/344dc25d39db7f897590a2855d45bd08a7d643c4/libraries/utils/generics/applicationlogger.cpp#L118 breaks compialtion on Win7. Do we even need this @RDoerfel ?

RDoerfel:
Fix #500 
- DWORD produced errors in Win7 and it is not used in Win10 either, so it is deleted. 
- test: minor changes
- example: use time vector instead of position file since this is the most likely usecase; Don't write error as gof in position file

### DIFF
--- a/examples/ex_hpiFit/main.cpp
+++ b/examples/ex_hpiFit/main.cpp
@@ -98,7 +98,7 @@ void write_pos(const float time, QSharedPointer<FIFFLIB::FiffInfo> info, Eigen::
         }
     }    
 
-    double error = std::accumulate(vGoF.begin(), vGoF.end(), .0) / vGoF.size();
+    // double error = std::accumulate(vGoF.begin(), vGoF.end(), .0) / vGoF.size();
     QQuaternion quatHPI = QQuaternion::fromRotationMatrix(rot);
 
     //qDebug() << "quatHPI.x() " << "quatHPI.y() " << "quatHPI.y() " << "trans x " << "trans y " << "trans z " << std::endl;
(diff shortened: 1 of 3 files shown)


examples/ex_hpiFit/main.cpp
```diff
@@ -156,17 +156,18 @@ int main(int argc, char *argv[])
...
-    // Read Quaternion File
-    Eigen::MatrixXd pos;
-    qInfo() << "Specify the path to your position file (.txt)";
-    UTILSLIB::IOUtils::read_eigen_matrix(pos, QCoreApplication::applicationDirPath() + "/mne-cpp-test-data/Result/ref_hpiFit_pos.txt");
-    RowVectorXd time = pos.col(0);
+    int N = ceil((last-first)/quantum);
+    RowVectorXf time(N);
+    for(int i = 0; i < N; i++){
+        time(i) = i * dT_sec;
+    }
```
LorenzE: You could try using `RowVectorXf::LinSpaced(N, 0, N-1) * dT_sec` for this, see Eigen docu [here](https://eigen.tuxfamily.org/dox/group__TutorialAdvancedInitialization.html).

libraries/utils/generics/applicationlogger.cpp
```diff
@@ -104,8 +104,6 @@ void ApplicationLogger::customLogWriter(QtMsgType type,
         // Enable colored output for
         // Set output mode to handle virtual terminal sequences
         HANDLE hOut = GetStdHandle(STD_OUTPUT_HANDLE);
-        DWORD dwMode = 0;
-        dwMode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
```
LorenzE: Was this tested on windows10 and linux?

RDoerfel: Tested on Win 10 and this shouldn't apply for Linux because of ifdef etc.

PR #509, opened 2020-03-02
[Fix] Small fixes in example wizard and FiffSimulator.cfg

RDoerfel:
Example wizard: 
- remove typo and add library
- remove return redundancy
- use QCoreApplication instead of QApplication to avoid error and:
> QCoreApplication - base class. Use it in command line applications.
QGuiApplication - base class + GUI capabilities. Use it in QML applications.
QApplication - base class + GUI + support for widgets. Use it in QtWidgets applications.

https://forum.qt.io/topic/94834/differences-between-qapplication-qguiappication-qcoreapplication-classes/3

FiffSimultor.cfg:
- use full path since it might be necessary on some operating systems (e.g. CentOs)

### DIFF
--- a/tools/wizards/mnecpp/example/main.cpp
+++ b/tools/wizards/mnecpp/example/main.cpp
@@ -67,8 +67,8 @@
  */
 int main(int argc, char *argv[])
 {
-    qInstallMessageHandler(ApplicationLogger::CustomLogWriter);
-    QApplication a(argc, argv);
+    qInstallMessageHandler(UTILSLIB::ApplicationLogger::customLogWriter);
+    QCoreApplication a(argc, argv);
 
     // Command Line Parser
     QCommandLineParser parser;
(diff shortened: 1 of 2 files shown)


tools/wizards/mnecpp/example/main.cpp
```diff
@@ -67,8 +67,8 @@
  */
```
LorenzE: Add `#include <utils/generics/applicationlogger.h>`

RDoerfel: https://github.com/mne-tools/mne-cpp/blob/d979ea1c1f66500fea7818454e8bd8f52fa2a4b5/tools/wizards/mnecpp/example/main.cpp#L38

I think it is already included?

LorenzE: sry did not see this. my bad.

PR #517, opened 2020-03-07
[DOC,ENH] Add Unit Test guide to website and Test Wizard

### LINKED ISSUE #496: Add docu for writing tests
We should add a docu Page under develop which shows how to create a test and add it to testframes sub project.

RDoerfel:
Thi PR includes following to mne-cpp and closes #496 :

- guide on how to write a unit test 
- templates for tests that can be included into QtCreator as wizard

Guide Can be seen here, let me know what you think.
https://rdoerfel.github.io/mne-cpp/pages/contribute/test.html
note: the website might not represent the actual status since  I experienced some issues with triggering a new update.


tools/wizards/mnecpp/test/wizard.json
```diff
@@ -0,0 +1,141 @@
...
+                    "type": "LineEdit",
+                    "data": {}
+                },	
+
+                {
+                    "name": "TestClassName",
+                    "trDisplayName": "Test Class Name:",
+                    "mandatory": true,
+                    "type": "LineEdit",
+                    "data": {}
```
LorenzE: Can we add default values for the class name such as `Test<YourTestClassName>` and for the project name `test_<YourTestName>`

RDoerfel: I could change it for the class name, but it seems that changing the file name is not straight forward.

PR #518, opened 2020-03-08
[FIX] Fix path for brainflow libs

Andrey1994:

### DIFF
--- a/.github/workflows/devbuilds.yml
+++ b/.github/workflows/devbuilds.yml
@@ -37,6 +37,7 @@ jobs:
         git push -f --tag
         # Create new dev_build release
         hub release create -m "Development Builds" dev_build --prerelease
+
   Linux:
     runs-on: ubuntu-16.04
     needs: UpdateDevTag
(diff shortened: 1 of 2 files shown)


Andrey1994:
wierd but it didnt solve the issue on linux

Andrey1994:
but Windows build is ok now

Andrey1994:
@LorenzE lets try PATH and LD_LIBRARY_PATH env vars?

LorenzE:
Either that or how about copying the lib files to the /lib instead of /bin folder? Also, I just realized that the MacOS changes I introduced did not work either. I will give the MacOS build another try later today.

Andrey1994:
<< how about copying the lib files to the /lib instead of /bin folder?
These libs should be in searchpath in runtime, the easiest option for it - place them to the same folder where executable is located and I think its bin folder. But I have no idea what linudeployqt does if it handles it then lets place to lib folder instead. Or copy to both of them

PR #523, opened 2020-03-14
[MAINT] Add Application Logger to tests and apply coding conventions

RDoerfel:
This PR includes:
- add Application Logger to all test frames
- add test_filtering to testframes.pro since it was somehow not there
- apply coding conventions to most of the .cpp files
- remove test_new_3d since it is not a test, not working and not in testframes.pro


RDoerfel:
@LorenzE do you have an idea why the linux test test_dipole_fit failed? It is running for Windows and Mac. I only added the application logger to this file.

LorenzE:
> @LorenzE do you have an idea why the linux test test_dipole_fit failed? It is running for Windows and Mac and I only added the application logger to this file.

Hmm don't know why the dipole fit test is failing. The dipole moment seems to be different on Linux. Didn't we have the same problem some weeks ago?

RDoerfel:
I think so, @LorenzE. Do you remember how you solved it?

LorenzE:
> I think so, @LorenzE. Do you remember how you solved it?

I vaguely remember the problem was solved after we did some changes to the CI. Can you try to remove all changes from the dipole test again and see if that changes anything?

RDoerfel:
> I vaguely remember the problem was solved after we did some changes to the CI. Can you try to remove all changes from the dipole test again and see if that changes anything?

@LorenzE Seems like it is working now, but I only added the application logger so no idea what the problem was.

PR #538, opened 2020-03-27
[FIX, MAINT] MNE Scan fixes and improvements

LorenzE:
This PR fixes/improves:

 - Connections are now loaded correctly from xml files on MNE Scan startup for the BabyMEG plugin.
 - Enforced some coding conventions in the BabyMEG plugin.
 - Hide the SQUID control widget action in the BabyMEG plugin since it is not used at the moment.
 - Fix the restarting of plugins
 - Remove loading of additional head points in the HpiSettingsView, since they are not needed at this point.
 - Make real-time covariance estimation work as a static function instead of a signal/slot communication.
 - Fix loading of stored setup parameters in the Covariance plugin.
 - Refactor forward solution loading and clustering in the RTC-MNE plugin.
 - Improve raw data visualization by not plotting sub-pixel wise in the RtFiffRawViewer.


libraries/disp/viewers/helpers/rtfiffrawviewdelegate.cpp
```diff
@@ -550,53 +535,61 @@ void RtFiffRawViewDelegate::createPlotPath(const QModelIndex &index,
...
+    QPointF qSamplePosition;
 
+    double dValueScaled, dValue;
+    double dScaleY = option.rect.height()/(2*dMaxValue);
     double y_base = path.currentPosition().y();
-    QPointF qSamplePosition;
 
-    double dDx = ((float)option.rect.width()) / t_pModel->getMaxSamples();
+    // Calculate the smallest possible width for one sample data point
+    int iSkip = t_pModel->getMaxSamples() / option.rect.width();
```
LorenzE: @gabrielbmotta Maybe we can do something similar in MNE Analyze?

gabrielbmotta: Yes, I tried doing something similar in mneanalyze,

```
    int iPaintStep = (int)(1.0/dDx) - 1;
    if (iPaintStep < 2){
        iPaintStep = 1;
    }

    for(int j = 0; j < data.size(); j = j + iPaintStep) {
        dValue = data[j] * dScaleY;

        //Reverse direction -> plot the right way
        newY = y_base - dValue;

        qSamplePosition.setY(newY);
        qSamplePosition.setX(path.currentPosition().x() + (dDx * (float)iPaintStep));
        path.lineTo(qSamplePosition);
    }
```

LorenzE: any problems with the scroll bar, spacers, etc?

gabrielbmotta: none yet. I also changed how the spacers were being drawn, before we were wasting time looping through spacers that weren't getting displayed. Now we only draw enough to fill the screen. It seems to be a bit smoother than before with these two changes, I'm still looking for 
ways to streamline it more.

PR #541, opened 2020-03-30
[WIP][ENH,MAINT] Accelerate HPI fit and add freq ordering to plugin 

RDoerfel:
This PR adds the frequency ordering to the hpi plugin in mne-scan and improves the performance of the hpi fitting. 

**frequency ordering:**
- add a push button to control panel and implement it into the hpi-plugin

**performance improvement:**
I was able to improve the performance of the hpi fit from **~250 ms** per fit to **~65 ms** per fit on my laptop. I got this improvement by:

- move calculation of everything that only has to be calculated once to the constructor (~225 ms)
- optimize parameters of minimization algorithm fminsearch (~112 ms)
- replace qlist in Sensor Struct with Eigen Matrices (~65 ms)

The result is shown in the following figure, just to give an impression. The values are calculated on one .fif file and all in all ~360 fits were performed. The figure shows the result of each improvement described above.
Figure_1

**Everything else**
- apply changes in the use of hpifit class everywhere where needed
- apply pedantic naming conventions

### DIFF
--- a/applications/mne_scan/plugins/hpi/FormFiles/hpisetupwidget.cpp
+++ b/applications/mne_scan/plugins/hpi/FormFiles/hpisetupwidget.cpp
@@ -59,8 +59,6 @@ HpiSetupWidget::HpiSetupWidget(Hpi* toolbox, QWidget *parent)
 , m_pHpi(toolbox)
 {
     ui.setupUi(this);
-
-    // connect(ui.m_qPushButton_About, SIGNAL(released()), this, SLOT(showAboutDialog()));
 }
 
 //=============================================================================================================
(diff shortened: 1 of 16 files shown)


RDoerfel:
[WIP] because I have to remove file path changes from ex_hpiFit once I'm sure there will be no further changes to make

applications/mne_scan/plugins/hpi/FormFiles/hpisetupwidget.cpp
```diff
@@ -60,7 +60,7 @@ HpiSetupWidget::HpiSetupWidget(Hpi* toolbox, QWidget *parent)
 {
     ui.setupUi(this);
 
-    connect(ui.m_qPushButton_About, SIGNAL(released()), this, SLOT(showAboutDialog()));
```
LorenzE: Please remove this line and also the about push button in `hpisetup.ui`

Issue #533, opened 2020-03-26
Make all MNE Scan plugin setup widgets more consistent


LorenzE:
The setup widgets for all MNE Scan plugins are shown when there is no measurement running. 

The setup widgets should only include: 
- All necessary setup widgets: block size, connection status, etc.
- A quick explanation about how to work with this plugin

The setup widgets should NOT include:
- Any logos of institutions/affiliations
- About widgets

LorenzE:
Closed by #563

PR #549, opened 2020-04-03
MNE Scan: fix splash screen messages and solve some compilation warnings 

juangpc:


applications/mne_scan/libs/scShared/Management/pluginmanager.h
```diff
@@ -154,6 +155,9 @@ class SCSHAREDSHARED_EXPORT PluginManager : public QPluginLoader
      */
     inline const QVector<IIO*>& getIOPlugins();
 
+Q_SIGNALS:
```
LorenzE, line 158: we use the `signals` over `Q_SIGNALS` syntax throughout the project

juangpc:
Hi, 

Quite a few files affected, but if you actually look at the code most of them are just explicit castings, member initialization ordering, and things like that. All changes deal with fix some small details with mne-scan.

I'd like to make some more comments:

 - Apart from the actual warning fixing, which are obviously related to different libraries linked during mne-scan compilation, the main thing in this pr is the messaging between mainwindow and the splash screen. That was not working and it seems to be working now. I had to divide the mainwindow constructor into several methods, so that I could allow for certain objects to exist, before using them. 

The solution I implemented is that being MainWindow the one that configures the application, etc... there is going to be a new member there to hold the splash screen address. That way I can setup communication between the pluginmanager and the splash screen. This was originally intended by the mne-scan author, but it never got to work. To have a splashscreen window which is refered or dependant from mainwindow, makes sense to me. After all, MainWindow class is the settings controller class that sets up pretty much everything.

 - After going through the code, I see how the model view controller pattern is not kept in this app. Being MainWindow the one that sets up everything... makes things a bit dependant on an actual window. I'm missing a true configurator class acting as an interface between the ui window and the actual app. 

- There are still a lot of warnings. I think we should slowly but Shirley (😝[:)](https://www.youtube.com/watch?v=ixljWVyPby0) ) start reducing them. Quite a few of them are just member initialization reordering and implicit castings.

- I see the way in which plugins are loaded... I think this is undesirable. To check and object_cast through a list of object-plugins just contradicts why we should all bother doing OOP in the first place. I get that it works and it is fast, after all it is all references annotation. But at the pluginmanager level, you shouldn't have to know about that. At least that's the way I see it. We should use dynamic binding of member methods to objects at run time. I think I can change this only adding some dispatcher class. But I don't want to do it unless you people also agree. Maybe this is related to [this](https://github.com/mne-tools/mne-cpp/issues/533#issue-588702855)? or maybe it is not... 

- More important... the application crashes when you try to exit it.

Thanks, and sorry for this long msg.

applications/mne_scan/mne_scan/main.cpp
```diff
@@ -122,18 +122,23 @@ int main(int argc, char *argv[])
 
     QPixmap pixmap(":/images/splashscreen.png");
     MainSplashScreen::SPtr splashscreen(new MainSplashScreen(pixmap));
+
+    MainWindow mainWin;
+    mainWin.setSplashScreen(*splashscreen);
```
LorenzE, line 127: Why did you choose to share the splash screen with the main window itself? Why not use a signal in MainWindow which forwards the plugin message to the splash screen living outside of MainWindow? This way the two do not need to know about each other.

applications/mne_scan/plugins/babymeg/babymeg.cpp
```diff
@@ -122,7 +122,7 @@ BabyMEG::~BabyMEG()
 QSharedPointer<IPlugin> BabyMEG::clone() const
 {
     QSharedPointer<BabyMEG> pBabyMEGClone(new BabyMEG());
-    return pBabyMEGClone;
+    return std::move(pBabyMEGClone);
```
LorenzE, line 125: I like this. Could you enforce this kind of behavior in all plugins' clone() methods defined in [plugins.pro](https://github.com/mne-tools/mne-cpp/blob/master/applications/mne_scan/plugins/plugins.pro)?

libraries/communication/rtClient/rtcmdclient.h
```diff
@@ -199,6 +199,7 @@ class COMMUNICATIONSHARED_EXPORT RtCmdClient : public QTcpSocket
     CommandManager  m_commandManager;   /**< The command manager. */
     QMutex          m_qMutex;           /**< Access serialization between threads */
     QString         m_sAvailableData;   /**< The last received response. */
+    using QTcpSocket::connectToHost;
```
LorenzE, line 202: why is this needed?

libraries/communication/rtClient/rtdataclient.h
```diff
@@ -136,7 +137,7 @@ class COMMUNICATIONSHARED_EXPORT RtDataClient : public QTcpSocket
 
 private:
     qint32 m_clientID;  /**< Corresponding client id of the data client at mne_rt_server */
-
+    using QTcpSocket::connectToHost;
```
LorenzE, line 140: why is this needed?

libraries/mne/mne_hemisphere.h
```diff
@@ -227,7 +227,7 @@ inline bool operator== (const MNEHemisphere &a, const MNEHemisphere &b)
 {
     if(a.pinfo.size() == b.pinfo.size()) {
         for(int i = 0; i < a.pinfo.size(); ++i) {
-            if(!a.pinfo.at(i).isApprox(b.pinfo.at(i), 0.0001f)) {
```
LorenzE, line 230: why was the precision for comparison removed?

libraries/utils/utils_global.h
```diff
@@ -49,7 +49,7 @@
 #if defined(STATICLIB)
 #  define UTILSSHARED_EXPORT
 #elif defined(UTILS_LIBRARY)
-#  define UTILSSHARED_EXPORT Q_DECL_EXPORT    /**< Q_DECL_EXPORT must be added to the declarations of symbols used when compiling a shared library. */
+#  define UTILSSHARED_EXPORT Q_DECL_EXPORT  __attribute__ ((visibility ("default")))  /**< Q_DECL_EXPORT must be added to the declarations of symbols used when compiling a shared library. */
```
LorenzE, line 52: I do not understand what is happening here. Can you explain pls?

applications/mne_scan/libs/scShared/Management/pluginmanager.h
juangpc: Ok. Just read they are just defines. I'll change to enforce a bit of homogeneity.

LorenzE: yes, they are just defines. this is only to be in sync with the rest of the project.

applications/mne_scan/mne_scan/main.cpp
juangpc: Because I don't want the pluginmanager to know anything about splashing windows. Because they live in different parts of my head? I believe that mainwindow is a bad name for a configuration interface class. There, it is ok to have knowledge of the pluginmanager (which is how it is done right now) and it seems ok to me, to have some reference to hold where the splash window is... should there be one.

applications/mne_scan/plugins/babymeg/babymeg.cpp
juangpc: Maybe. jk

libraries/communication/rtClient/rtcmdclient.h
juangpc: becuase the signature is different, and the compiler warns you that you are actually not overriding.

libraries/communication/rtClient/rtdataclient.h
juangpc: why as in why do we all need mne-cpp? jk
please see previous comment.

libraries/mne/mne_hemisphere.h
juangpc: because it is not actually being used. the precision is casted to ze-ro. so i figured that if it hasn't been used all along... then we're sure we can live without it. Apart, I'm not sure 100%, but I think dose eigen people have somekind of precission implemented in their isapprox thing.

libraries/utils/utils_global.h
juangpc: visibility while linking libraries... what a wonderful subject to read at 1am...
https://gcc.gnu.org/wiki/Visibility

libraries/mne/mne_hemisphere.h
LorenzE: https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#title33

libraries/communication/rtClient/rtcmdclient.h
LorenzE: Can you share the exact warning please and also where it is shown? I assume https://github.com/mne-tools/mne-cpp/blob/98f9741f82d70fd489882ad2d73f8af0eeee5678/libraries/communication/rtClient/rtcmdclient.cpp#L75

applications/mne_scan/mne_scan/main.cpp
LorenzE: I aggree. However, with my proposed idea the PluginManager would not have to know about the SplashScreen if you would forward the signal from the pluginmanager through the mainwindow to the main method. Only main() would know about the splash screen.

I do not like the fact that the splash screen is now shared between main() and the MainWindow. Why does the MainWindow need to know about the splash screen at all? And if even if this is absolutley necessary why not putting the splash screen directly into the MainWindow then?

LorenzE:
> * After going through the code, I see how the model view controller pattern is not kept in this app. Being MainWindow the one that sets up everything... makes things a bit dependant on an actual window. I'm missing a true configurator class acting as an interface between the ui window and the actual app.

Yes, in this case the MainWindow provides the main view and and at the same time setups everything around it. However, I do not see any data models in this setup so I would argue a MVC pattern was not anticipated here.

> * I see the way in which plugins are loaded... I think this is undesirable. To check and object_cast through a list of object-plugins just contradicts why we should all bother doing OOP in the first place. I get that it works and it is fast, after all it is all references annotation. But at the pluginmanager level, you shouldn't have to know about that. At least that's the way I see it. We should use dynamic binding of member methods to objects at run time. I think I can change this only adding some dispatcher class. But I don't want to do it unless you people also agree. Maybe this is related to [this](https://github.com/mne-tools/mne-cpp/issues/533#issue-588702855)? or maybe it is not...

The upcast has to happen somewhere because further down the line we do rely on the information whether the loaded plugin is a sensor or algorithm plugin. The upcast ist performed after loading the plugin, so that the developer has the plugins already sorted into a sensor and algorithm plugin list, which he/she can use to access the loaded plugin more conveniently.

> * More important... the application crashes when you try to exit it.

Can you provide more information? What kind of plugins were added to the pipeline? Was the pipeline running when you closed MNE Scan? One way to get a good idea of what is happening is to take a look at the call stack:
1) start the mne_rt_server
2) Start MNE Scan in Release mode **via the QtCreator** by clicking on the play button with the little bug on it, see pic below. 

image

If MNE Scan crashes you can check the caller stack in the QtCreator

image

applications/mne_scan/mne_scan/main.cpp
juangpc: I like your idea of keeping splash screen in the MainWindow class. I actually didn't do that just because I didn't want to change the code that much. I think that would actually make a lot of sense.

>However, with my proposed idea the PluginManager would not have to know about the SplashScreen ... 

I honestly forgot about what changes we did online. I tried to find them... but I couldn't, so I started again. But if you want to do it, it will be a nice exercise for me. 
The configuring class (i.e. MainWindow), configures and uses the plugin manager, and if there is a splash screen, connects it to that thing.
PluginManager just broadcasts. That's the idea behind signals, right? PluginManager doesn't know about who's listening. Where do you see PluginManager acknowledging the splash screen?

>Only main() would know about the splash screen.

But that way, why would main need to know if there is such a thing called plugin? It really doesn't belong there.

> I do not like the fact that the splash screen is now shared between main() and the MainWindow. 

First of all, it is only a reference, which can be not configured. Secondly, I agree 100%. 😄 

>if you would forward the signal from the pluginmanager through the mainwindow to the main method

Trying to understand what you mean. Do you mean to have MainWindow act as a relay of signals? But why would you want to do that? Signals are a broadcasting mechanism. Very cool. You want to listen somewhere?... you just tune in. What if you want to add some log class with some slot to which you also want to connect the plugin manager? You can do that now, independently of everything else.

So would you fork and maybe show me? Or through a skype (very trendy these days)?
We could focus on this and the static compilation.

libraries/communication/rtClient/rtcmdclient.h
juangpc: image

libraries/mne/mne_hemisphere.h
juangpc: Sorry, I'm a bit dense, right now. Could you explain?

Following your previous comment, let me show you the actual warning that motivated this change.

image

juangpc:
>I would argue a MVC pattern was not anticipated here

I don't know Lorenz, maybee I'm wrong... Let me ask this question i've learned while learning c++: Would mne-cpp work ok without a gui? 
If the answer is no, then the mvc has problems. What I see is that a class called MainWindow does all the work. Apart from the name, it would be hard to separate both functionalities. One the windows and other the actual application itself doing stuff. 

> The upcast has to happen somewhere

Completely agree. But it should happen where I think it is supposed to happen in oop languages.. under the curtain. The programmer should not be distracted by that. Not with an if-else structure. That's how I see it. These kinds of procedures, I think, are hard and I'm still getting used to them, but the main thing is to battle complexity. If the application's complexity grows... I think we will miss them.  So let the casting happen somewhere where c++ compiler developers wanted it to happen. I. think this is one of the very few things I learned during 2019. And it was thanks to you. Honestly I think this is important.

Regarding the call stack. I didn't have the rt server running when I saw the crash. I'll try to replicate
Here is some report.
image

And the file with the 
[crash.txt](https://github.com/mne-tools/mne-cpp/files/4429814/crash.txt)

libraries/mne/mne_hemisphere.h
LorenzE: ah sry `pinfo` is of type `Eigen::VectorXi`. In this case you are right, we are dealing with an integer value and the precision can be discarded.

applications/mne_scan/mne_scan/main.cpp
LorenzE: yes,  let's discuss this in a Skype.

libraries/communication/rtClient/rtcmdclient.h
LorenzE: thx. Now I understand. so what happens is that the name of this function was chosen poorly since it is identical with the one in QAbstractSocket and thus tries to overload it. Honestly, I do not understand how `using QTcpSocket::connectToHost;` gets rid of the warning, but it does 

We try not to use `using` statements in headers. This [article](https://stackoverflow.com/questions/5849457/using-namespace-in-c-headers) explains why that is.

How about if we overload the function properly?

```
/**
 * Connect to a mne_rt_server using port 4217
 *
 * @param[in] hostName      The hostName on the given port.
 * @param[in] port          The port.
 * @param[in] openMode      The open mode. Default is QIODevice::OpenMode::ReadWrite.
 * @param[in] protocol      The protocol. Default is QAbstractSocket::NetworkLayerProtocol::AnyIPProtocol.
 */
void connectToHost(const QString &hostName,
                   quint16 port,
                   QIODevice::OpenMode openMode = ReadWrite,
                   QAbstractSocket::NetworkLayerProtocol protocol = AnyIPProtocol);
```

```

void RtCmdClient::connectToHost(const QString &hostName,
                                quint16 port,
                                QIODevice::OpenMode openMode,
                                QAbstractSocket::NetworkLayerProtocol protocol)
{
    Q_UNUSED(port)
    QTcpSocket::connectToHost(hostName, 4217, openMode, protocol);
}
```

libraries/communication/rtClient/rtdataclient.h
LorenzE: see answer above

LorenzE:
> Would mne-cpp work ok without a gui?

Parts of MNE-CPP work without a GUI, MNE Scan does not. MNE Scan is not build based on a MVC pattern. At least not the part we are discussing about. It has some flavors of it, for example how it loads the plugins and assigns the appropriate visualization corresponding to the plugin's output data. There are some MVC patterns in MNE Scan, e.g., the real-time raw data display, butterfly view etc. They follow a MVC pattern. MNE Scan as the base application (without the plugins, visualization etc.) does not follow MVC. However, I aggree that MainWindow is somewhat in an in-between state, providing a view and at the same time controlling a lot of the plugin setups. Although it is not meant to follow a clean MVC pattern, a controller class would clean up the structure a bit. Always room for improvements ;) Maybe you want to take a look and work out a controller class?

> The programmer should not be distracted by that. Not with an if-else structure.

At some point the programmer will have to know whether this IPlugin instance he/she is currently working with is either a sensor or an algorithm plugin. At that point an upcast becomes necesary followed by an if to check if the pointer is valid. I do understand what you are getting at, I think it is just more of a question where you want to perform the upcast.

libraries/communication/rtClient/rtcmdclient.h
juangpc: What about this call?
https://github.com/mne-tools/mne-cpp/blob/eb75b90aa61a9ccedbfd2cfaed82ff38b330284f/applications/mne_scan/plugins/fiffsimulator/fiffsimulator.cpp#L285

I don't really like: ``` m_pRtCmdClient->connectToHost(m_sFiffSimulatorIP,0);  ```

If we define a pointer to a qthread object, we can call its method, and our lives would be much happier.

libraries/communication/rtClient/rtdataclient.h
juangpc: happier life above.

juangpc:
> At some point the programmer will have to know whether this IPlugin instance he/she is currently working with is either a sensor or an algorithm plugin. At that point, an upcast becomes necessary followed by an if to check if the pointer is valid.

After some offline conversation, I think I was able to express a bit better what I meant. Just in case someone else is reading. This is what I mean. You can do it with a dispatcher (visitor) class or with a set of different methods inside the actual plugin manager class. Either way, it is true there is going to be some object casting, however, that will be done by the actual c++ language, not by us.

```c++
foreach(QString file, PluginsDir.entryList(QDir::Files))
{
    this->setFileName(PluginsDir.absoluteFilePath(file));
    IPlugin *pPlugin = this->instance();
    if(pPlugin)
    {
        PluginDispatcher pd;
        pPlugin->acceptDispatch(pd);   --> this method sends a reference of this to 
                                       --> the dispatch methodin PluginDispatcher.    
    }
}

class PluginDispatcher {
public:
   void dispatch(PluginClassA& p) //-> do what you know you need to do with ClassA
   void dispatch(PluginClassB& p) //-> do what you know you need to do with ClassB
};


```

libraries/communication/rtClient/rtcmdclient.h
LorenzE: > define a pointer to a qthread object, we can call its method

What do you mean by that exactly? 

> I don't really like: m_pRtCmdClient->connectToHost(m_sFiffSimulatorIP,0);

Where does the second parameter (`0`) come from?

How about we set the `port ` to default to 4217? This not also gets rif of the warning but also opens up the possibility to connect to a different port sinc ethe port parameter is now actually use.

```
/**
 * Connect to a mne_rt_server using port 4217 as default.
 *
 * @param[in] hostName      The hostName on the given port.
 * @param[in] port          The port. Default is 4217.
 * @param[in] openMode      The open mode. Default is QIODevice::OpenMode::ReadWrite.
 * @param[in] protocol      The protocol. Default is QAbstractSocket::NetworkLayerProtocol::AnyIPProtocol.
 */
void connectToHost(const QString &hostName,
                   quint16 port = 4217,
                   QIODevice::OpenMode openMode = ReadWrite,
                   QAbstractSocket::NetworkLayerProtocol protocol = AnyIPProtocol);
```

```

void RtCmdClient::connectToHost(const QString &hostName,
                                quint16 port,
                                QIODevice::OpenMode openMode,
                                QAbstractSocket::NetworkLayerProtocol protocol)
{
    QTcpSocket::connectToHost(hostName, port, openMode, protocol);
}
```

juangpc: OK, even better. I like this so much more. I thought you wanted to keep the ingrained port number.

juangpc:
I've done a lot of changes in a lot of files... It was a little bit of a wild ride, navigating through all that c-ish code out there... But fun. I think I know the project a bit better now. 
Some of the suggestions by @LorenzE were not met, only because I wanted to have a compiling version without warnings. So please, if you see fit, re-suggest, and I'll be glad to address changes.
Besides, the splash screen thing. It is now working. Maybe we can get rid of the splash presence in main. Now that it is all working, I can do that, or let it like it is right now...

applications/mne_scan/libs/scShared/Management/pluginconnectorconnectionwidget.cpp
```diff
@@ -117,7 +117,10 @@ PluginConnectorConnectionWidget::PluginConnectorConnectionWidget(PluginConnector
 
     //Connect Signals
     foreach(QComboBox* m_pComboBox, m_qMapSenderToReceiverConnections)
-        connect(m_pComboBox, static_cast<void (QComboBox::*)(const QString &)>(&QComboBox::currentIndexChanged), this, &PluginConnectorConnectionWidget::updateReceiver);
+        connect(m_pComboBox,
```
LorenzE, line 120: We usually put sender and receiver info in separate lines. Please change to:

```
connect(m_pComboBox, static_cast<void (QComboBox::*)(const QString &)(&QComboBox::currentTextChanged),
        this, &PluginConnectorConnectionWidget::updateReceiver);
```

applications/mne_scan/mne_scan/main.cpp
```diff
@@ -122,18 +122,23 @@ int main(int argc, char *argv[])
...
-        int p = (i*100)/time;
-        splashscreen->showMessage("Loading modules.."+ QString::number(p)+"%");
-    }
+//    //ToDo Debug Some waiting stuff to see splash screen -> remove this in final release
+//    int time = 100;
+//    for(int i=0; i < time;++i)
+//    {
+//        int p = (i*100)/time;
+//        splashscreen->showMessage("Loading modules.."+ QString::number(p)+"%");
+//    }
```
LorenzE, line 137: the for loop can be removed

applications/mne_scan/mne_scan/mainwindow.cpp
```diff
@@ -113,8 +113,39 @@ MainWindow::MainWindow(QWidget *parent)
...
+    if(m_pToolBar) {
+        if(m_pLabelTime) {
+            delete m_pLabelTime;
+        }
+        delete m_pToolBar;
+    }
+
+    delete m_pDynamicPluginToolBar;
+}
+
```
LorenzE, line 135: add spacer `//=============================================================================================================`

applications/mne_scan/mne_scan/mainwindow.cpp
```diff
@@ -113,8 +113,39 @@ MainWindow::MainWindow(QWidget *parent)
...
+void MainWindow::setupPlugins()
+{
+    if(m_pSplashScreen.data())
+    {
+        QObject::connect(m_pPluginManager.data(),&PluginManager::pluginLoaded,
+                         m_pSplashScreen,&MainSplashScreen::showMessage);
+    }
     m_pPluginManager->loadPlugins(qApp->applicationDirPath()+pluginDir);
```
LorenzE, line 144: add spacer

applications/mne_scan/mne_scan/mainwindow.cpp
```diff
@@ -134,24 +165,10 @@ MainWindow::MainWindow(QWidget *parent)
     initStatusBar();
 }
```
LorenzE, line 167: add spacer

libraries/communication/rtClient/rtcmdclient.cpp
```diff
@@ -68,13 +68,17 @@ RtCmdClient::RtCmdClient(QObject *parent) :
             &RtCmdClient::sendCommandJSON);
 }
 
-//=============================================================================================================
-
-void RtCmdClient::connectToHost(QString &p_sRtServerHostName)
-{
-    QTcpSocket::connectToHost(p_sRtServerHostName, 4217);
-}
+////=============================================================================================================
```
LorenzE, line 71: This commented function can be removed

libraries/communication/rtClient/rtcmdclient.h
```diff
@@ -86,16 +86,28 @@ class COMMUNICATIONSHARED_EXPORT RtCmdClient : public QTcpSocket
...
-    /**
-     * Connect to a mne_rt_server using port 4217
-     *
-     * @param[in] p_sRtServerHostName    The IP address of the mne_rt_server
-     */
-    void connectToHost(QString &p_sRtServerHostName);
-
+    explicit RtCmdClient(QObject *parent = Q_NULLPTR);
+
+//    //=========================================================================================================
```
LorenzE, line 91: commented functions can be removed

libraries/communication/rtClient/rtdataclient.h
```diff
@@ -82,16 +82,29 @@ class COMMUNICATIONSHARED_EXPORT RtDataClient : public QTcpSocket
...
 
     //=========================================================================================================
-    /**
-     * Connect to a mne_rt_server using port 4218
-     *
-     * @param[in] p_sRtServerHostName    The IP address of the mne_rt_server
-     */
-    void connectToHost(const QString& p_sRtServerHostName);
-
+//    /**
```
LorenzE, line 88: can be removed

libraries/utils/utils_global.h
LorenzE: I would like to put a hold on this change until I familiarize myself with what it exactly does. Also, if we decidde to add this, we should try changing it project wide. Maybe this is something for a future PR?

juangpc:
If you disable ftbuffer plugin in mne_scan. There is no crash on exit.

libraries/utils/utils_global.h
juangpc: I think it just optimizes the content of a dll so that only the used functions are prepared. And this way the load times are much shorter.

juangpc:
Very nice. 
I would only comment that the signal-> slot connection is now on the splash screen side. Depending on how you see it, it could be 50% in the receiver (splash screen) or in the emitter (plugin manager). I think that I would see it as more obvious to lay in the emitter since it is the one actually emitting. But, if you see it as something that belongs in the splash screen I don't have too hard feelings about it.

PR #551, opened 2020-04-03
[ENH] Add advanced model for hpi fit

RDoerfel:
This PR adds an advanced model for the sinoides by adding line frequency and harmonics, linear componend and DC componend . This is implemented as extra mode, so you have to choose it when initializing the constructor. Per default, the fast and basic model is used.

### DIFF
--- a/libraries/inverse/hpiFit/hpifit.cpp
+++ b/libraries/inverse/hpiFit/hpifit.cpp
@@ -83,14 +83,14 @@ using namespace FWDLIB;
 
 HPIFit::HPIFit(FiffInfo::SPtr pFiffInfo, bool bDoFastFit)
     : m_bDoFastFit(bDoFastFit)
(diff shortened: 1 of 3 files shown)


libraries/inverse/hpiFit/hpifit.cpp
```diff
@@ -81,21 +81,23 @@ using namespace FWDLIB;
 // DEFINE MEMBER METHODS
 //=============================================================================================================
 
-HPIFit::HPIFit(FiffInfo::SPtr pFiffInfo)
+HPIFit::HPIFit(FiffInfo::SPtr pFiffInfo, bool bDoFastFit)
```
LorenzE, line 84: put funtion parameters underneath each other

RDoerfel:
> Found some coding convention mismatches after all ;)

haha when you think you got everything but then more coding conventions come arround the corner

PR #555, opened 2020-04-08
[FIX] Fix FiffSimulator crashes

LorenzE:
Fix crashes relating to the FiffSimulator plugin in MNE Scan introduced by initializing some pointer too late.

### DIFF
--- a/applications/mne_scan/plugins/fiffsimulator/fiffsimulatorproducer.cpp
+++ b/applications/mne_scan/plugins/fiffsimulator/fiffsimulatorproducer.cpp
@@ -146,7 +146,7 @@ void FiffSimulatorProducer::run()
     connectDataClient(m_pFiffSimulator->m_sFiffSimulatorIP);
 
     qint32 count = 0;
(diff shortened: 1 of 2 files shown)


applications/mne_scan/plugins/fiffsimulator/fiffsimulatorproducer.cpp
```diff
@@ -146,7 +146,7 @@ void FiffSimulatorProducer::run()
     connectDataClient(m_pFiffSimulator->m_sFiffSimulatorIP);
 
     qint32 count = 0;
-    while(m_pRtDataClient->state() != QTcpSocket::ConnectedState) {
+    while(!isInterruptionRequested() && m_pRtDataClient->state() != QTcpSocket::ConnectedState) {
```
juangpc, line 149: While correct, operator precedence hacking is a bit too much hack-stylee. Would suggest use of perenthesis to make explicit what is implicit by the compiler rules.

imsorryk:
mne-rt-notworking

Still seems to have an issue connecting

PR #556, opened 2020-04-08
[MAINT] Minor changes fiff info

juangpc:
QList uses this shared memory mechanism by Qt where objects are shared and updated only on changes. So there is no need to do explicit deep copy. 

I'm not sure if the copy constructor is needed at all.


libraries/fiff/fiff_info_base.h
```diff
@@ -216,7 +216,7 @@ class FIFFSHARED_EXPORT FiffInfoBase
     QString filename;           /**< Filename when the info is read of a fiff file. */
     QStringList bads;           /**< List of bad channels. */
     FiffId meas_id;             /**< Measurement ID. */
-    fiff_int_t  nchan;          /**< Number of channels. */
+    fiff_long_t  nchan;          /**< Number of channels. */
```
LorenzE, line 219: why the change to 64bit precision? `nchan` should be int with 32bit precision. https://github.com/mne-tools/fiff-constants/blob/6733c5eb8a3e036c83ab10915de487fe77016546/DictionaryTags.txt#L183

juangpc: I think that should be the way to go. Making mne-cpp more high-level by slowly abstracting such intricate differences and selecting things that will just not bother you (-> you can abstract them).  
The reason for this particular case is that this value will get promoted later to double-precision because it will be compared with an output that is just that. But that is this particular case. I think that the code should become less c-ish and just don't get on the way of further dev.
But au, I don't have a very strong opinion on that. So whatever you guys decide.

LorenzE: So why don't fix the comparison instead of the actual precision, which is fixed by the Fiff file format standard?

juangpc: Ok. ok. I see you don't like it. 🤣So I'll give you two answers: 
-Because the standard/library needs to update into a thing-that-is-not-in-the way, and this is a minuscule step towards that direction.
-I'll fix the comparison.

LorenzE: It is not that I do not like it. It is that `nchan` is int32 by fiff definition. Changing it to 64bit is not messing with the fiff specification. It also breaks the build, see checks below, because we do not have a pre defined function to write int64 to fiff file.

>Because the standard/library needs to update into a thing-that-is-not-in-the way, and this is a minuscule step towards that direction.

I do not see how a 32/64bit precicion change results in a library with fewer things-in-the-way. I do understand that changing code which compares between different types incorrectly needs to be fixed. Maybe I am missing the point here?

juangpc: what can I say? It's true. What you say, makes sense to me. A lot of sense.
I can't find the original comparison that triggered the idiocy. but I fixed the change.

libraries/fiff/fiff_info.cpp
```diff
@@ -322,28 +322,28 @@ FiffInfo FiffInfo::pick_info(const RowVectorXi &sel) const
...
-    res.nchan  = sel.size();
+    res.nchan  = static_cast<int>(sel.size());
 
     return res;
 }
 
 //=============================================================================================================
 
-QList<FiffChInfo> FiffInfo::set_current_comp(QList<FiffChInfo>& chs, fiff_int_t value)
+QList<FiffChInfo> FiffInfo::set_current_comp(QList<FiffChInfo>& _chs, fiff_int_t value)
```
LorenzE, line 332: Please have a look at our [coding conventions](https://mne-cpp.github.io/pages/contribute/conv_style.html). If you want to change this parameter's name I would propose `lChs`.

juangpc: done.

libraries/fiff/fiff_info_base.h
juangpc: done

PR #561, opened 2020-04-10
[DOCU] Update WASM tutorial

LorenzE:
Update WASM tutorial with Qt 5.15 info and fix git clone command mistake.


juangpc:
OK!
I was able to follow all the steps in the tutorial and didn't notice the error. 😞

PR #562, opened 2020-04-14
[ENH] Update brainflow version to add unicorn

Andrey1994:

new release of brainflow supports [unicorn device](https://www.unicorn-bi.com/)(for now only on linux)


Andrey1994:
I am just curios about libbluetooth3, I dont use any system packages and there are no system level dependencies in BrainFlow... Some third party SDKs require BLUEZ but I dont link BrainFlow with these libs and call dlopen manually instead if device which requires this SDK was selected.

Andrey1994:
```
ERROR: ldd outputLine: "libbluetooth.so.3 => not found"
ERROR: for binary: "/home/runner/work/mne-cpp/mne-cpp/applications/mne_scan/plugins/brainflowboard/brainflow/java-package/brainflow/src/main/resources/libunicorn.so"
ERROR: Please ensure that all libraries can be found by ldd. Aborting.
```
I dont link BrainFlow with libunicorn.so it looks like linuxdeployqt runs ldd for all libraries/binaries in a folder even if they are not used. I didnt know that

PR #580, opened 2020-04-27
[ENH,MAINT] Update Fwd. Solution with new Head Position 

RDoerfel:
This PR aims to make head position updates to the forward solution possible. Therefore I had to introduce a couple of changes.

1. Split computeForward into:
- initFwd() that contains initialization, loading and preparation of source space, coil set and bem model
- calculateFwd() - actual calculation of the MEG and EEG forward solution
- storeFwd() - write solution to file

2. Make the most important variables a member for ComputeFwd()
- ToDo: once refactored MNEForwardSolution should be the main output that stores the solution for further use

3. Add function updateHeadPos() that takes a meg->head transformation and recalculates the meg forward solution for updated coil set
- ToDo: once refactored it might be possible to recalculate a bit less

4. Update test_mne_forward_solution

5. Add example that shows the newly introduced workflow 
- The Workflow is going to change a bit once the FWDLIB is successfully refactored. So far we have to store the fwd solution and read it with MNEForwardSolution to create this object. In the future, computeForward should provide MNEForwardSolution as a member.

### DIFF
--- a/applications/mne_forward_solution/main.cpp
+++ b/applications/mne_forward_solution/main.cpp
@@ -76,6 +76,7 @@ int main(int argc, char *argv[])
     ComputeFwdSettings settings(&argc,argv);
     ComputeFwd cmpFwd(&settings);
     cmpFwd.calculateFwd();
+    cmpFwd.storeFwd();
 
     return app.exec();
 }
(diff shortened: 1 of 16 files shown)


RDoerfel:
With the last commit I replaced the MneNamedMatrix that stores the lead field Matrix with FiffNamedMatrix

PR #595, opened 2020-05-06
[WP][ENH,FIX,MAINT] MNE Anonymize 

juangpc:
Hi, 

Well, this is the update I mentioned before. In the process of creating the gui, I've inevitably found some bugs and some inconsistencies between the controller and the anonymizer class. There are a big number of changes, but these are not substantial. The application pretty much does the same thing although I think it all is much better structured, more oo, and easier to follow (I hope). The list of changes worth mentioning are: 

bug fixes. 
1. measurement date and birth date options were not working before. A problem with the way in which the qdate class reads the formated input string. 
2. some files don't follow strictly the fiff standard and contain a list of tags where the last tag in the list is not marked with a NEXT=-1. Thus, the reader fell through the end of the file. The main loop now controls for that too.
3. member anonymizer class is now stored in the heap. This doesn't qualify as a bug, strictly speaking, but could end up causing problems if the class grows too much in the future. (And I think it really makes more sense to be there. It is not a small object).
4. the opening of the input stream called the open method of fiff_stream which does some business itself which caused some repeating of the verification of the first tags. We now use the open member method of the QIOdevice. 
5. This is definitely not a bug, other than perhaps a bug in the developer. Everything related with the tag directory was not working. The anonymizer performed some business in relation to the tag directory but the final write-down of the directory was commented. There was a bug in there which got fixed. The problem is that once you have a directory you need to update the pointer to the directory tag (101). This update was also not working at all. I tried to make it work but I couldn't... until I figured it out. The problem was that the stream does not let you go back and write in a random position of a file. You would need to open another stream or another fileio... just to update the pointer. Since the whole thing was really not needed at all. I have eliminated all the business related to the tag directory. It is not used, not updated and not saved in the output file.

Enhancements.
1. The app is no single-threaded. After testing speed, we all agree it is not really needed. This reduces considerably its weight and its complexity. (don't know if this is strictly an enhancement?)
2. Instead of having an input and output tag for the reading and output, everything is now performed with a single tag. This tag is now stored in the heap, and located with a member pointer. So much more robust in case a big tag is processed. Besides, it makes the application 12% faster on my mac. 😄 
3. Some changes accommodate the idea of different controllers for gui and cl apps. All using the same anonymizer model. Following then the MVC pattern as described in the docs of MNE-CPP.
4. Quite a bit of refactoring, but most importantly a more thoughtful oo approach, where local stack vars are promoted to member vars, whenever they are heavily used and shared between methods. This makes the code so much readable. For instance, the main loop of the application was 
```c++
    while(pInTag->next != -1) {
        inStream.read_tag(pInTag);

        updateBlockTypeList(pInTag);
        censorTag(pOutTag,pInTag);

        // The order of the tags in the output file is sequential. No jumps in the output file.
        if(pOutTag->next > 0) {
            pOutTag->next = FIFFV_NEXT_SEQ;
        }
        addEntryToDir(pOutTag,outStream.device()->pos());
        FiffTag::convert_tag_data(pOutTag,FIFFV_NATIVE_ENDIAN,FIFFV_BIG_ENDIAN);
        outStream.write_tag(pOutTag);
    }
```

and is now 

```c++
    while( (m_pTag->next != -1) && (!m_pInStream->device()->atEnd()))
    {
        readTag();
        censorTag();
        writeTag();
    }
```
5. Some unused methods have been cleaned. I intend to copy them back in the gui version. But for now, I want to see which ones are finally necessary. 
6. Quite some documentation improvements. Both in the actual comments which are later parsed doxygen style, and in the help info shown to the user. 
7. Instead of having the --no-gui option for command line and the default mode to be a gui. I figure that it is always easier to have the user not remember to include the --no-gui option, and at the same time, it is quite easy to make a gui shortcut launcher to include the --gui flag.
8. There are these members which, if others think so, I think could be promoted to the fiff library. 
```inline QString FiffAnonymizer::subjectHandToString(const int handCode) const```
```inline QString FiffAnonymizer::subjectSexToString(const int sexCode) const```
There are others, like data to string for instance... (we can discuss this in another thread).
9. main now uses qcoreapplication's and qapplications event loop. Previously since this was just a command line thing, we were returning 0 from main. Now we use the canonical ```return app->exec();``` from qt, even for the command line application. This made me understand the buffered signaling capabilities of qcoreapplication. This will be necessary when we create the gui version of the app.
10. Some warning in the fiff library have been addressed.
11. A bit more systematic approach to location of bugs and see the overal behaviour has been adopted. I've tested the following list of options, sometimes many times in a row, to test for stability in the results. This way I've been able to locate quite a few bugs. 
This was before I decided to change the ```--no-gui``` for a ```--gui``` one. I think it is evident how it seems more practical this new way. 

--version --no-gui
--version --no-gui -i one.fif
--no-gui --verbose -i one.fif
--no-gui --verbose -i six.fif
--no-gui --verbose -o six.fif
--no-gui --verbose -i bla
--no-gui --verbose -i five.fif -o bla
--no-gui --verbose -i five.fif -o seven.fif
--no-gui --verbose -i five.fif
--no-gui --verbose -i five.fif -o five.fif
--no-gui --verbose -i five.fif -o bla --brute
--no-gui -i six.fif
--no-gui  -o six.fif
--no-gui -i bla
--no-gui -i five.fif -o bla
--no-gui -i five.fif -o seven.fif
--no-gui -i five.fif
--no-gui -i five.fif -o five.fif
--no-gui -i five.fif -o bla --brute
--no-gui --verbose -i five.fif -o ua.fif--delete_input_file_after
--no-gui --verbose -i four.fif -o four.fif--delete_input_file_after --avoid_delete_confirmation
--no-gui --verbose -i five.fif --measurement_day 20171002 
--no-gui --verbose -i five.fif --measurement_day 20171002 --measurement_day_offset 3
--no-gui --verbose -i five.fif  --measurement_day_offset 3
--no-gui --verbose -i five.fif --subject_birthday 20171002 
--no-gui --verbose -i five.fif --subject_birthday 20171002 --subject_birthday_offset 3
--no-gui --verbose -i five.fif --subject_birthday 20171002 --subject_birthday_offset 3
--no-gui --verbose -i five.fif --measurement_day_offset 3 --his 4

12. And lastly, maybe a bit controversial and highly personal decision. I think I need to explain my train of thought better. The use of ```std::printf``` instead of ```qInfo() ``` when interacting with the user comes from various ideas. 
First, I've noticed that qInfo, qDebug, qWarning, qCritical, and so on, can be deleted during compilation with specific compilation flags. I think that these are intended for debugging and/or specialized development purposes. But the menus like "Do you want to delete this file? [Y/n]" I think are not part of this objective, these are part of the main interaction with the user (given it is a command-line application). If you delete these messages during compilation you have an incomplete app.

Second, I find the "[INFO]" label at the start of every message distracting and also self-explicatory. From the user's perspective, who doesn't know (nor needs to know) between the different qInfo, qDebug etc... it means. "Here is some info" followed by some info... Which is quite redundant. 
For instance, the qcommandlineparser class, has a way to automatically generate an output help message to the user. This text doesn't show any [INFO] label. And typically, it is uncommon to see this kind of message when using command line applications. At least that is my experience. 
That is why I decided to use ```std::printf``` for user interaction.

This is how it looks with ```std::printf```
image

This is how it looks with ```qInfo()```:

image

And here an example of when the -v option is used (verbose mode)
image

image

Thanks for getting to this point.

### CODE REVIEW

File: testframes/test_mne_anonymize/test_mne_anonymize.pro [hunk 1]
```diff
@@ -42,7 +42,6 @@ QT += testlib concurrent network
 QT -= gui
 
 CONFIG   += console
-CONFIG   -= app_bundle
```
LorenzE, line 45: The `CONFIG   -= app_bundle` flag needs to be set. Tests are not supposed to be .app bundles. This should fix the current build.

applications/mne_anonymize/mne_anonymize.pro
```diff
@@ -44,7 +44,7 @@ QT += widgets concurrent network
 
 VERSION = $${MNE_CPP_VERSION}
 
-CONFIG   += console
+CONFIG   += consol
```
LorenzE, line 47: Typo. Should be `console`

testframes/test_mne_anonymize/test_mne_anonymize.pro
juangpc: aham .... ok. I'm so used to mess with the flag to be able to debug on mac that I forgot about it.

applications/mne_anonymize/mne_anonymize.pro
juangpc: I'm the typo

applications/mne_anonymize/fiffanonymizer.cpp
```diff
@@ -716,282 +227,445 @@ int FiffAnonymizer::censorTag(FiffTag::SPtr outTag,
...
+        printIfVerbose("Warning: Beware that a subject''s face can be reconstructed from it");
+        printIfVerbose("Warning: This software can not anonymize MRI data, at the moment.");
+        printIfVerbose("  ");
         break;
     }
     default:
+    {   }
+    }
+}
+
```
LorenzE, line 484: add spacers

applications/mne_anonymize/fiffanonymizer.cpp
```diff
@@ -716,282 +227,445 @@ int FiffAnonymizer::censorTag(FiffTag::SPtr outTag,
...
+{
+    static QStringList subjectHandRefList = { "unknown", "right", "left", "ambidextrous" };
+    if ((handCode > -1) && (handCode < subjectHandRefList.size()))
+    {
+        return subjectHandRefList.at(handCode);
+    } else {
+        qCritical() << "Invalid subject handedness code. [0 = unknown, 1 = right, 2 = left, 3 = ambidextrous]. The code of the subject is: " << QString::number(handCode);
+        return QString("invalid-code");
+    }
+}
```
LorenzE, line 507: add newline

applications/mne_anonymize/fiffanonymizer.cpp
```diff
@@ -716,282 +227,445 @@ int FiffAnonymizer::censorTag(FiffTag::SPtr outTag,
...
-    m_fFileOut.setFileName(m_sFileNameOut);
+QString FiffAnonymizer::getFileNameOut() const
+{
+    return m_fFileOut.fileName();
 }
 
 //=============================================================================================================
 
-void FiffAnonymizer::setVerboseMode(bool bFlag)
+void FiffAnonymizer::setVerboseMode(const bool bFlag)
```
LorenzE, line 661: `const` is not needed here since the passed value is not a reference

applications/mne_anonymize/fiffanonymizer.cpp
```diff
@@ -1005,118 +679,133 @@ void FiffAnonymizer::setBruteMode(bool bFlag)
...
 {
-    m_bUseSubjectBirthdayOffset = true;
-    m_iSubjectBirthdayOffset = iSubjBirthdayOffset;
+    return m_dMeasurementDate;
 }
 
 //=============================================================================================================
 
-void FiffAnonymizer::setDeleteInputFileAfter(bool bFlag)
+void FiffAnonymizer::setMeasurementDayOffset(const int iMeasDayOffset)
```
LorenzE, line 718: `const` is not needed here since the passed value is not a reference

applications/mne_anonymize/fiffanonymizer.cpp
```diff
@@ -1005,118 +679,133 @@ void FiffAnonymizer::setBruteMode(bool bFlag)
...
-    return randomFileName.append(".fif");
+QDateTime  FiffAnonymizer::getSubjectBirthday()
+{
+    return m_dSubjectBirthday;
 }
 
 //=============================================================================================================
 
-void FiffAnonymizer::deleteInputFile()
+void FiffAnonymizer::setSubjectBirthdayOffset(const int iSubjBirthdayOffset)
```
LorenzE, line 747: `const` is not needed here since the passed value is not a reference

applications/mne_anonymize/fiffanonymizer.h
```diff
@@ -60,7 +62,7 @@
 #include <QDateTime>
 #include <QDebug>
 #include <QPointer>
-
+#include <QString>
 //=============================================================================================================
```
LorenzE, line 66: add new line

applications/mne_anonymize/fiffanonymizer.h
```diff
@@ -124,134 +126,133 @@ class FiffAnonymizer
...
     /**
-     * If found in the fiff file, the specified number of days will be subtracted from the measurement date information contained in each fif file.
+     * Retrieves the state of the FiffAnonyzer object's desired verbose mode.
+     *
+     * @param [out] Bool value with the actual verbose mode.
      *
-     * @param [in] iMeasDayOffset   Integer with the number of dates to subtract from the measurement date.
      */
-    void setMeasurementDayOffset(int iMeasDayOffset);
+    bool getVerboseMode() const;
```
LorenzE, line 189: see coment in function defintion

applications/mne_anonymize/fiffanonymizer.h
```diff
@@ -124,134 +126,133 @@ class FiffAnonymizer
...
+     * Configure the state of the FiffAnonymizer object's desired anonymization mode. If set to TRUE, apart from the
+     * default information additional information will also be anonymized, like Subject's weight, height, sex or
+     * handedness and the project's information too.
+     *
+     * @param [in] bFlag    Bool argument whether to use the brute mode.
      *
-     * @param [in] sSubjBirtday String containing the desired subject birthday date.
      */
-    void setSubjectBirthday(const QString& sSubjBirthday);
+    void setBruteMode(const bool bFlag);
```
LorenzE, line 200: see coment in function defintion

applications/mne_anonymize/fiffanonymizer.h
```diff
@@ -261,161 +262,74 @@ class FiffAnonymizer
...
 
     //=========================================================================================================
     /**
-     * Value of User's request to avoid confirmation prompt for input file deletion.
+     * If found in the fiff file, the specified number of days will be subtracted from the subject's birthday date information contained in each fif file.
+     *
+     * @param [in] iSubjBirthdayOffset  Integer with the number of dates to subtract from the subject's birthday date.
      */
-    bool getDeleteInputFileConfirmation();
+    void setSubjectBirthdayOffset(const int iSubjBirthdayOffset);
```
LorenzE, line 291: see coment in function defintion

applications/mne_anonymize/fiffanonymizer.h
```diff
@@ -440,80 +354,30 @@ class FiffAnonymizer
...
-     * This method allows to transform between a vector and a regular tag structure. The vector will only contain
-     * the data inside the Tag Directory tag (remember a tag directory is a tag itself). This allows to use all the
-     * methods related to a tag (i.e. writeTag) while storing a newly updated tag directory into a fiff file.
-     *
-     * @param [in] pTag A pointer to a tag where the tag directory information can be stored.
+     * This transforms the data inside a subject sex tag (male, female or unknown) into a string object
+     * containing that information.
      */
-    void dir2tag(FIFFLIB::FiffTag::SPtr pTag);
+    inline QString subjectSexToString(const int sexCode) const;
```
LorenzE, line 372: see coment about `const` in function defintion

applications/mne_anonymize/fiffanonymizer.h
```diff
@@ -440,80 +354,30 @@ class FiffAnonymizer
...
-     * @param [out] stream  Output stream
-     * @param [in] tagKind  Tag kind code to search for
-     * @param [in] newPos   New position to store in the data field of the pointer tag.
+     * This transforms the data inside a subject handedness tag (right, left or unknown) into a string object
+     * containing that information.
      */
-    void updatePointer(FIFFLIB::FiffStream* stream,
-                       FIFFLIB::fiff_int_t tagKind,
-                       FIFFLIB::fiff_long_t newPos);
+    inline QString subjectHandToString(const int handCode) const;
```
LorenzE, line 380: see coment about `const` in function defintion

applications/mne_anonymize/fiffanonymizer.h
```diff
@@ -525,108 +389,105 @@ class FiffAnonymizer
...
-    int m_iDfltSubjectSex;              /**< Subject's sex substitutor.*/
-    int m_iDfltSubjectId;               /**< Subject's id substitutor.*/
-    int m_iDfltSubjectHand;             /**< Subject's hand substitutor.*/
-    int m_iSubjectWeight;               /**< Subject's weight substitutor.*/
-    int m_iSubjectHeight;               /**< Subject's height substitutor.*/
-    int m_iProjectId;                   /**< Project's id# substitutor.*/
+    FIFFLIB::FiffStream::SPtr m_pInStream;   /**< Pointer to FiffStream object for reading.*/
+    FIFFLIB::FiffStream::SPtr m_pOutStream;  /**< Pointer to FiffStream object for writing the result.*/
+    FIFFLIB::FiffTag::SPtr m_pTag;         /**< Pointer to FiffTag used for reading and writing each tag.*/
+    //FIFFLIB::FiffTag::SPtr m_pOutTag;        /**< Pointer to FiffTag used for writing each anonymized tag.*/
```
LorenzE, line 471: dow we still need this? otherwise plear remove commented line

applications/mne_anonymize/main.cpp
```diff
@@ -3,10 +3,10 @@
  * @file     main.cpp
  * @author   Juan Garcia-Prieto <juangpc@gmail.com>;
  *           Lorenz Esch <lesch@mgh.harvard.edu>;
- *           Matti Hamalainen <msh@nmr.mgh.harvard.edu>;
  *           Wayne Mead <wayne.mead@uth.tmc.edu>;
- *           John C. Mosher <John.C.Mosher@uth.tmc.edu>
- * @since    0.1.0
+ *           John C. Mosher <John.C.Mosher@uth.tmc.edu>;
+ *           Matti Hamalainen <msh@nmr.mgh.harvard.edu>;
+ * @version  dev
```
LorenzE, line 9: `@version` has been replaced with `@since`

applications/mne_anonymize/main.cpp
```diff
@@ -40,8 +40,8 @@
 // INCLUDES
 //=============================================================================================================
 
-#include "settingscontroller.h"
-
+#include "settingscontrollerCL.h"
```
LorenzE, line 43: `settingscontrollerCL.h` should be `settingscontrollercl.h` captial letters in file names are discouraged.

applications/mne_anonymize/mainwindow.h
```diff
@@ -0,0 +1,17 @@
+#ifndef MAINWINDOW_H
```
LorenzE, line 1: please make this follow our class template

applications/mne_anonymize/mne_anonymize.pro
```diff
@@ -69,12 +69,14 @@ CONFIG(debug, debug|release) {
...
+    fiffanonymizer.cpp \
+    mainwindow.cpp \
+    settingscontrollerCL.cpp
 
 HEADERS  += \
-    settingscontroller.h \
-    fiffanonymizer.h
+    fiffanonymizer.h \
+    mainwindow.h \
+    settingscontrollerCL.h
```
LorenzE, line 79: `settingscontrollerCL.h` should be `settingscontrollercl.h` captial letters in file names are discouraged.

applications/mne_anonymize/mne_anonymize.pro
```diff
@@ -69,12 +69,14 @@ CONFIG(debug, debug|release) {
 DESTDIR = $${MNE_BINARY_DIR}
 
 SOURCES += main.cpp \
-    settingscontroller.cpp \
-    fiffanonymizer.cpp
+    fiffanonymizer.cpp \
+    mainwindow.cpp \
+    settingscontrollerCL.cpp
```
LorenzE, line 74: `settingscontrollerCL.cpp` should be `settingscontrollercl.cpp` captial letters in file names are discouraged.

applications/mne_anonymize/settingscontrollerCL.cpp
```diff
@@ -0,0 +1,536 @@
...
+
+//=============================================================================================================
+// DEFINE GLOBAL METHODS
+//=============================================================================================================
+
+//=============================================================================================================
+// DEFINE MEMBER METHODS
+//=============================================================================================================
+
+SettingsControllerCL::SettingsControllerCL(const QStringList& arguments)
```
LorenzE, line 70: class names should be camel case `SettingsControllerCL` -> `SettingsControllerCl`

applications/mne_anonymize/settingscontrollerCL.h
```diff
@@ -0,0 +1,243 @@
...
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ *
+ * @brief     SettingsControllerCL class declaration.
+ *
+ */
+
+#ifndef MNEANONYMIZE_SettingsControllerCL_H
```
LorenzE, line 39: `MNEANONYMIZE_SETTINGSCONTROLLERCL_H`

applications/mne_anonymize/settingscontrollerCL.h
```diff
@@ -0,0 +1,243 @@
...
+     * Prints a message if the Verbose option has been set.
+     */
+    void printIfVerbose(const QString& str) const;
+
+signals:
+    //=========================================================================================================
+    /**
+     * Signal the main qt core application to stop running and exit.
+     */
+    void finished(const int) const;
```
LorenzE, line 205: see `const` comments above

applications/mne_anonymize/settingscontrollerCL.h
```diff
@@ -0,0 +1,243 @@
...
+
+    //=========================================================================================================
```
LorenzE, line 207: remove this `//======================`

applications/mne_anonymize/mainwindow.h
juangpc: I ve deleted the whole class in order to create it from scratch. I'll make sure the new class will use the template.

applications/mne_anonymize/settingscontrollerCL.h
juangpc: done.

applications/mne_anonymize/settingscontrollerCL.h
juangpc: done.

applications/mne_anonymize/settingscontrollerCL.cpp
juangpc: Sorry, I don't understand.

applications/mne_anonymize/fiffanonymizer.cpp
juangpc: I think you mean to unpack those lists. Done that.

applications/mne_anonymize/fiffanonymizer.cpp
juangpc: I understand, but won't it be of help to the compiler? Again, this is just a stupid trivial program, but I'm trying to learn to do things properly. I won't modify the value, so it will help the compiler (I think).

applications/mne_anonymize/fiffanonymizer.cpp
juangpc: Rethinking it again, and reading about it. I've come to the conclusion that it is better to not use the const modifier. It sort-of-breaks my dear encapsulation. I like that. Thanks Lorenz.

applications/mne_anonymize/main.cpp
juangpc: done

applications/mne_anonymize/main.cpp
juangpc: done

applications/mne_anonymize/mne_anonymize.pro
juangpc: done

applications/mne_anonymize/mne_anonymize.pro
juangpc: done

applications/mne_anonymize/settingscontrollerCL.cpp
juangpc: I see now. l is not L

applications/mne_anonymize/fiffanonymizer.cpp
LorenzE: yes it might help the compiler optimize, although this is highly questionable with the function at hand. This discussion might be an interesting read: https://stackoverflow.com/questions/117293/use-of-const-for-function-parameters

applications/mne_anonymize/fiffanonymizer.cpp
LorenzE: no I mean a new line before the `//====================` spacer

applications/mne_anonymize/fiffanonymizer.cpp
LorenzE: `//====================` missing

applications/mne_anonymize/fiffanonymizer.h
LorenzE: ?

applications/mne_anonymize/settingscontrollercl.cpp
```diff
@@ -0,0 +1,574 @@
...
+//    if(execute())
+//    {
+//        qCritical() << "Error during the anonymization of the input file";
+//        emit finished(1);
+//        return;
+//    }
+
+//    emit finished(0);
+}
+
```
LorenzE, line 106: add spacer

Issue #593, opened 2020-05-05
MNE Analyze plugins are not included on mac dynamic builds


LorenzE:
MNE Analyze plugins are not included into the bundle on mac dynamic builds. This is most likely a bug in the mne_analyze.pro file when deploying the application. Thx @agramfort for reporting this.

PR #599, opened 2020-05-06
[Fix] Fix mne analyze plugin deployment on macos

LorenzE:
Fix copying of mne analyze plugins on macos. #593


juangpc:
When it finishes I'll download and test.

LorenzE:
> When it finishes I'll download and test.

Thx!

PR #616, opened 2020-05-14
[ENH,MAINT] Add (real-time) Forward Solution Plugin to MNE Scan

### LINKED ISSUE #467: RTC-MNE plugin crashes
RTC-MNE plugin crashes when deleted during clustering.

### LINKED ISSUE #557: MNE Scan: Quick Control HPI Fitting
elements in tab HPI Fitting do not scale with widget

RDoerfel:
This small PR adds a real-time forward solution plugin to MNE Scan. The plugin calculates the measurement specific part of the forward solution and provides this as its output. It is possible to include continuous hpi information and update the forward solution to the new head position. To implement the new plugin into the Mne Scan workflow, quite a view changes to other plugins where necessary. Affected Plugins are: 

**HPI - Plugin:**

- Improve control widget and split it into tabs (Load Digitizers, Fitting, Head Movements) which closes #557 
- Introduce Head Movement detection (movement in mm and rotation in degree) and thresholds for each to define big head movements.
- implement RealTimeHpiRsult into PluginConnectorConnection to make hpi results accessible

**Source Localization (RTCMNE):**

- move all forward solution related tasks to new rtFwd plugin
- load mri-head transformation to visualize average head (Hpi-Plugin) and Brain surface in same space
- add mri-head transformation to RealTimeSourceEstimate

The new plugin introduces the following workflows and possibilities:

**Whats new:**

- forward solution calculation for EEG and MEG in separate plugin
- head position update (connection to hpi plugin)
- clustering 

**What does it need:**

- source space
- bem model
- measurement file (for compensator information, also if they are not in use)

**Connections/Workflow:**

- Sensor Plugins (Input)
- Hpi Plugin (Input)
- Source Localization (Output)

The Plugin can be connected to all sensor plugins that provide a RealTimeMultSampmleArray as output or RealTimeHpiResult.  The necessary part is the FiffInfo. Hence it might be possible to connect to other plugins as well. As output, the  Plugin provides a RealTimeForwardSolution, which provides a FiffInfo, MNEForwardSolution, and the lead field matrices. 

**What's next:**

The Plugin is not yet real-time capable. This hopefully changes soon. A head position update with followed clustering takes around 4-5 seconds at the moment. Since adult persons don't move that much per measurement, this might be still sufficient. But of course, there is room for improvement.

Further, the new workflows have to be documented on our website. This will be addressed in a future PR.

**ToDo in this PR:**
- [x] Mne Scan closes once a measurement is stopped, @LorenzE do you have any suggestions?

Closes #467

**Some pictures**

Control widget Forward Solution
fwdsol

Avarage Head + SensorSurface + Brain and head movement tab of hpi widget
headmove

Possible workflow with continious hpi, forward solution and source localization + Forward Solution Settings widget.
workflow


applications/mne_scan/plugins/rtfwd/rtfwd.cpp
```diff
@@ -163,6 +163,8 @@ void RtFwd::unload()
 
 bool RtFwd::start()
 {
+    // Maybe we can move all of this to the run() method?
```
RDoerfel, line 166: The idea on having that in start() was to check if all minimum necessary filepaths are set. You can't set them during run() so you should check before the measurement is started.

PR #632, opened 2020-05-27
[MAINT] Bump CI to Qt 5.15.0

LorenzE:
This PR bumps up the maximum allowed Qt version to Qt 5.15. Some changes to the pullrequest.yml Github Workflow were needed in order to deal with the fact that Qt 5.15 only comes pre built for MSVC 2019 on windows. Our static pre built binaries were updated to Qt 5.15 as well. The build guide was updated accordingly. Some how Qt3D's renderer plugins do not get copied via windployqt, so we have to copy them manually. 

WIP:

- [ ] Tests are failing for MSVC 2019 Qt 5.15. This does not happen locally on my machine. Any ideas @RDoerfel ?

### DIFF
--- a/.github/workflows/buildqtbinaries.yml
+++ b/.github/workflows/buildqtbinaries.yml
@@ -71,7 +71,7 @@ jobs:
         path: qt5_5150_static_binaries_macos.tar.gz
 
   GenerateWinStaticBinaries:
-    runs-on: windows-2016
+    runs-on: windows-2019
 
     steps:
     - name: Clone repository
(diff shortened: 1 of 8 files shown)


RDoerfel:
@LorenzE, I just built the project with Qt5.15 and it is working on my machine too. So I have no idea why it is failing, but I was able to find the [last call](https://github.com/mne-tools/mne-cpp/blob/0245b40adeb164d95ec2343fc3661876e384b8dd/libraries/fwd/computeFwd/compute_fwd.cpp#L2008) which ends [here](https://github.com/mne-tools/mne-cpp/blob/0245b40adeb164d95ec2343fc3661876e384b8dd/libraries/fiff/c/fiff_coord_trans_old.cpp#L613). I hope this helps.

LorenzE:
This needs some more CI testing. Closing for now.

PR #647, opened 2020-06-19
[MAINT,DOC,FIX] Minor changes and fixes to MNE Analyze

gabrielbmotta:
- corrected saving/loading multiview by passing settings path
- adjusted minimum window size
- changed signal view control widget scaling to allow for window sizes
- averaging plugin documentation page
- closing main window now calls destructors for views in dockwidgets
- fix multiple loading of buttons and tabs in averaging plugin when loading new / switching between files


applications/mne_analyze/mne_analyze/mainwindow.cpp
```diff
@@ -439,6 +442,8 @@ void MainWindow::createPluginViews(QSharedPointer<PluginManager> pPluginManager)
                 pWindow = m_pMultiView->addWidgetTop(pView, sCurPluginName);
             }
 
+            pWindow->setObjectName(sCurPluginName);
```
LorenzE, line 445: maybe we can move this function call to addWidetBottom and addWidgetTop? This way others wouldd not forget to set it when using the MultiView.

applications/mne_analyze/mne_analyze/mainwindow.cpp
```diff
@@ -419,9 +420,11 @@ void MainWindow::createPluginControls(QSharedPointer<ANSHAREDLIB::PluginManager>
 void MainWindow::createPluginViews(QSharedPointer<PluginManager> pPluginManager)
 {
     m_pGridLayout = new QGridLayout(this);
-    m_pMultiView = new MultiView();
+    m_pMultiView = new MultiView(m_sSettingsPath);
+    m_pMultiView->setParent(this);
```
LorenzE, line 424: is this needed? Doesn't the setCentralWidget call below already takes care of parenting the new multi view widget

applications/mne_analyze/mne_analyze/mainwindow.cpp
gabrielbmotta: Yes. Turns out we had already done that and i did not notice. I'll remove this as it is redundant.

applications/mne_analyze/mne_analyze/analyzecore.cpp
```diff
@@ -169,7 +169,7 @@ void AnalyzeCore::initMainWindow()
 {
     m_pMainWindow = new MainWindow(m_pPluginManager);
     QObject::connect(m_pMainWindow.data(), &MainWindow::mainWindowClosed,
-                     this, &AnalyzeCore::onMainWindowClosed);
+                     this, &AnalyzeCore::onMainWindowClosed, Qt::DirectConnection);
```
LorenzE, line 172: Why not AutoConnection? Are the two objects running in different threads?

applications/mne_analyze/mne_analyze/mainwindow.cpp
```diff
@@ -85,8 +85,9 @@ MainWindow::MainWindow(QSharedPointer<ANSHAREDLIB::PluginManager> pPluginManager
 , m_sSettingsPath("MNEANALYZE/MainWindow")
 , m_sCurrentStyle("default")
 {
+    this->setObjectName("mainwindow");
     setWindowState(Qt::WindowMaximized);
-    setMinimumSize(400, 400);
+    setMinimumSize(1280, 720);
```
LorenzE, line 90: Maybe use 800x450? just in case someone with a very small (and old) laptop uses MNE Analyze ;)

applications/mne_analyze/mne_analyze/mainwindow.cpp
```diff
@@ -129,7 +128,16 @@ void MainWindow::closeEvent(QCloseEvent *event)
...
+    for(QDockWidget* widget : this->findChildren<QDockWidget*>()){
+        widget->setAttribute(Qt::WA_DeleteOnClose);
+        widget->close();
+    }
+
     emit mainWindowClosed();
+    QMainWindow::closeEvent(event);
 
     // default implementation does this, so its probably a good idea
     event->accept();
```
LorenzE, line 143: I think we do not need this call anymore since we call `QMainWindow::closeEvent(event);` now

applications/mne_analyze/mne_analyze/mainwindow.cpp
```diff
@@ -390,10 +398,13 @@ void MainWindow::createPluginControls(QSharedPointer<ANSHAREDLIB::PluginManager>
     setTabPosition(Qt::RightDockWidgetArea,QTabWidget::East);
     setDockOptions(QMainWindow::ForceTabbedDocks);
 
+    pPluginManager->setParent(this);
```
LorenzE, line 401: PluginManager is part of AnalyzeCore where it is held as a QSharedPointer. Setting the parent here might interfere with QSahredPointer's automated garbage collection. Do we need to set the parent here?

applications/mne_analyze/mne_analyze/mainwindow.cpp
```diff
@@ -390,10 +398,13 @@ void MainWindow::createPluginControls(QSharedPointer<ANSHAREDLIB::PluginManager>
...
     //Add Plugin controls to the MainWindow
     for(IPlugin* pPlugin : pPluginManager->getPlugins()) {
         if(QDockWidget* pControl = pPlugin->getControl()) {
             addDockWidget(Qt::LeftDockWidgetArea, pControl);
+            pControl->setParent(this);
```
LorenzE, line 406: Do we need this if we use the `setAttribute(Qt::WA_DeleteOnClose)` solution from above?

libraries/disp/viewers/multiview.cpp
```diff
@@ -120,6 +120,8 @@ MultiViewWindow* MultiView::addWidgetBottom(QWidget* pWidget,
     pDockWidget->setWindowTitle(sName);
     pDockWidget->setWidget(pWidget);
 
+    pWidget->setParent(pDockWidget);
```
LorenzE, line 123: This is interesting. I did some reading. pDockWidget's setWidget function does not take ownership of pWidget. So setting the parent here makes sense. Does this alone fix the destructor calling problem? Or do we still need to set the Qt::WA_DeleteOnClose attribute?

applications/mne_analyze/mne_analyze/analyzecore.cpp
gabrielbmotta: I was testing to see if the inconsistent destructor calling was due to the analyzecore onWindowClosed being called at different times. It wasn't, and I forgot to change it back. Will remove.

PR #648, opened 2020-06-22
[ENH,MAINT] Write filtered data to file, delete data via GUI, remove minimal version mode

LorenzE:
This PR includes:

- Add support for writing filtered data to file in MNE Analyze
- Add support to delete models from the data manager in MNE Analyze
- Remove minimal version flag. This mode was needed to be able to build on the older Neuromag systems. Since we switched to the fieldtrip buffer, this mode is no longer needed.


applications/mne_analyze/libs/anShared/Management/analyzedata.cpp
```diff
@@ -161,22 +161,47 @@ QStandardItemModel* AnalyzeData::getDataModel()
...
+                    }
+                } else {
+                    if(m_pData->removeRows(index.row(), 1, index.parent())) {
+                        if(m_pData->rowCount() == 0) {
+                            emit modelIsEmpty();
+                        }
+                        qDebug() << "[AnalyzeData::removeModel] Removed model at index" << index;
+                        return true;
+                    } else {
+                        qDebug() << "[AnalyzeData::removeModel] Could not remove model at index" << index;
```
gabrielbmotta, line 198: Remove/replace qDebugs

applications/mne_analyze/plugins/datamanager/FormFiles/datamanagerview.cpp
```diff
@@ -124,3 +128,18 @@ void DataManagerView::onNewFileLoaded(int iSubject,
...
+
+//=============================================================================================================
+
+void DataManagerView::keyPressEvent(QKeyEvent *event)
+{
+    switch (event->key())
+    {
+        case Qt::Key_Delete:
+            emit removeItem(m_pUi->m_pTreeView->currentIndex());
+            break;
```
gabrielbmotta, line 140: When hitting the delete key (or right clicking and selecting delete) on the subject itself, the dialog pops up as if a model were selected. Is this fine as is for now or should there be some check to make sure the item being deleted is a file to allow the drop down/pop up to appear?
ts

LorenzE: good catch. I will fix this.

PR #650, opened 2020-06-25
[ENH] MNE Anonymize with GUI 🥳  and Webassembly ready 🏄🏻‍♂️

juangpc:
Hi,

Several features: 

 ``` ~$ mne_anonymize --help``` this will show all the options for the mne_anonymize command-line app.
With ``` ~$ mne_anonymize --gui``` a GUI is started. But other options included in the command line will help set up the GUI with the specified options. The nice part of it is that it is all done with the exact same code (yes, that means that a bug on the command-line will occur too in the GUI version... but I guess that's a good thing?).

future things needed... 

 - perhaps a modified documentation page (right now is all Command-line).
 - an icon for the application. Or at least a generic mne-cpp  one.
 - deployment in windows should include an application shortcut with the "mne_anonymize --gui" text in it.
perhaps someone knows how to detect if an application is being called from a command line or from a explorer window.
 - wasm works by writing an in.fif file into the private wasm drive, and an out.fif file will be generated during the anonymization. The problem is that in order to be able to import and download the app both files need to be duplicated (not at the same time). Thus a maximum of 4GB + 2GB + footprint of the application itself, thus around 6GB of ram would be required for the application to run OK with maximum sized fif files (2GB). That is not happening any time soon. So, wasm is only going to work for 1.3GB at most... for now.
Still I need to change the wasm compilation script so that wasm accepts 4GB total memory... right now it is set to 1GB. By default total memory is set to grow dynamically up to a maximum of 4GB. But for multithread, it is downsized to 1GB by default. So.... either we have a separate wasm compilation script for mne_analyze and mne_anonymize (first multithread, 2nd single threaded and memory growing dynamically). Or we modify the specific total memory maximum for mne_anonymize. I would need help for this. 

This was the best exercise ever... Thanks everybody.


applications/mne_anonymize/mne_anonymize.pro
```diff
@@ -40,15 +40,41 @@ include(../../mne-cpp.pri)
...
 contains(MNECPP_CONFIG, static) {
     CONFIG += static
     DEFINES += STATICBUILD
 }
 
+contains(MNECPP_CONFIG, wasm) {
+#    QMAKE_LFLAGS += -s ERROR_ON_UNDEFINED_SYMBOLS=1
+#    QMAKE_LFLAGS += -s ASSERTIONS=1
+#    QMAKE_LFLAGS += -s STRICT=0
+#    QMAKE_LFLAGS += -s FORCE_FILESYSTEM=1
```
LorenzE, line 67: try setting `QMAKE_LFLAGS += -s TOTAL_MEMORY=4GB` here. Afterwards you can try pushing your changes to your remote via `git push origin mne_anonymize_GUI:wasm`. The command will create a new wasm branch on your remote, if not already existing. This will trigger the CI to build the wasm version and host it on your gh-pages. Once the CI finishes you should be able to access your application via [https://svdecomposer.github.io/mne-cpp/mne_anonymize.html](https://svdecomposer.github.io/mne-cpp/mne_anonymize.html). This way we can check which web browser actually support the 4GB WASM memory flag.

juangpc: Hi, 
 - I'm trying to understand what do you mean by "here". I've moved the arg definitions upwards with the rest of them. Not sure that's what you meant. I've commented because I've tried a few things and it wasn't quite working. I didn't want to spend that much time on this at that point. 

 - Managed to get the wasm working locally. Both qt and mne-cpp compiling just fine in darwin!!! ✌🏻
 - Modified a bit the inner workings. Now you should be able to anonymize two files of ± 1.99GB or so (4gb - the footprint of the application which shouldn't be that much). The only con is that every time you anonymize a file you need to re-load because immediately after anonymization finishes, the **internal** copy of the input file gets deleted.

LorenzE: ah just saw you tried the flag below already. sry for the noise. were you able to see the flag int the cmd line output when compiling with emscripten? did it error during compilation?

applications/mne_anonymize/mne_anonymize.pro
```diff
@@ -40,15 +40,41 @@ include(../../mne-cpp.pri)
 
 TEMPLATE = app
 
-QT += widgets concurrent network
+QT += core gui widgets
```
LorenzE, line 43: Removing the `network` module leads to linkage problems in windows static builds. 

Also, by default `Qt` includes `gui` and `core`. So no need to set them here again. You can safely use `QT += widgets network`

juangpc: will do

applications/mne_anonymize/mainwindow.cpp
```diff
@@ -0,0 +1,925 @@
...
+        event->ignore();
+    }
+}
+
+//=============================================================================================================
+
+bool MainWindow::confirmClose()
+{
+    return true;
+//    const QMessageBox::StandardButton ret
```
LorenzE, line 116: do we still need this switch, otherwise remove

juangpc: At some point there could be some interest in: 

 - be able to modify each and every field.
 - if so, it could be interesting to save values in a text file that could be loaded back. If modifications are made, I want to be warned to save them before exiting. Code is just a reminder.

LorenzE: ok in that case we should keep it

applications/mne_anonymize/mainwindow.cpp
```diff
@@ -0,0 +1,925 @@
+//=============================================================================================================
+/**
+ * @file     mainwindow.cpp
+ * @author   Wayne Mead <wayne.mead@uth.tmc.edu>;
+ *           Juan Garcia-Prieto <juangpc@gmail.com>;
+ *           Lorenz Esch <lesch@mgh.harvard.edu>;
+ *           Matti Hamalainen <msh@nmr.mgh.harvard.edu>;
+ *           John C. Mosher <John.C.Mosher@uth.tmc.edu>
+ * @since    0.1.0
```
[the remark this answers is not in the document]
juangpc: That we do

applications/mne_anonymize/mainwindow.cpp
```diff
@@ -0,0 +1,925 @@
...
+ *
+ */
+
+//=============================================================================================================
+// INCLUDES
+//=============================================================================================================
+
+#include "mainwindow.h"
+#include "ui_mainwindow.h"
+#include <settingscontrollergui.h>
```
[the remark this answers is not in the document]
juangpc: done

LorenzE:
Can you add a new button for `MNE Anonymize - WASM` to the homepage's Release page? https://mne-cpp.github.io/pages/install/binaries.html Just like we have one for `MNE Analyze - WASM`

juangpc:
> Can you add a new button for `MNE Anonymize - WASM` to the homepage's Release page? https://mne-cpp.github.io/pages/install/binaries.html Just like we have one for `MNE Analyze - WASM`

https://svdecomposer.github.io/mne-cpp/pages/install/binaries.html

LorenzE:
> Like this?

I would prefer another button right next to the MNE Analyze one. This way we can avoid creating a new row in the table.

juangpc:
> > Like this?
> 
> I would prefer another button right next to the MNE Analyze one. This way we can avoid creating a new row in the table.

done.

PR #660, opened 2020-07-07
[ENH] Remove unnecesarry data copies when filtering

LorenzE:
Remove some deep copies when filtering in order to speed up filtering

### DIFF
--- a/libraries/rtprocessing/helpers/filterkernel.cpp
+++ b/libraries/rtprocessing/helpers/filterkernel.cpp
@@ -75,80 +75,84 @@ using namespace UTILSLIB;
 // DEFINE GLOBAL RTPROCESSINGLIB METHODS
 //=============================================================================================================
 
-QString RTPROCESSINGLIB::getStringForDesignMethod(const FilterKernel::DesignMethod &designMethod)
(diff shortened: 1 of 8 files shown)


libraries/rtprocessing/helpers/filterkernel.cpp
```diff
@@ -71,6 +71,88 @@ using namespace RTPROCESSINGLIB;
 using namespace Eigen;
 using namespace UTILSLIB;
 
+//=============================================================================================================
+// DEFINE GLOBAL RTPROCESSINGLIB METHODS
+//=============================================================================================================
+
+QString RTPROCESSINGLIB::getStringForDesignMethod(const FilterKernel::DesignMethod &designMethod)
+{
+    QString designMethodString = "External";
```
gabrielbmotta, line 80: Would these if statements be better as if-elseif or a switch statement w/ default to avoid having to check the input again once the output is assigned, and having to assign a value twice in case the initial assignment is overwritten? (same for the 3 following functions)

LorenzE: +1 good idea

PR #665, opened 2020-07-21
[FIX] MNE Analyze: fixed indexing when updating event group list.

gabrielbmotta:
Fixed issue where function was trying to get the name from an invalid index from iterating over map while not looking up the list of keys before reading from the map. Also changed return of function from QString& to QString for best practices.

### DIFF
--- a/applications/mne_analyze/libs/anShared/Model/annotationmodel.cpp
+++ b/applications/mne_analyze/libs/anShared/Model/annotationmodel.cpp
@@ -864,15 +864,27 @@ void AnnotationModel::setGroupName(int iGroupIndex,
 
 //=============================================================================================================
 
-const QString& AnnotationModel::getGroupName(int iGroupIndex)
+QString AnnotationModel::getGroupName(int iGroupIndex)
 {
     if(!m_mAnnotationHub.contains(iGroupIndex)){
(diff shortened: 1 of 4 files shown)


applications/mne_analyze/libs/anShared/Model/annotationmodel.cpp
```diff
@@ -864,17 +864,29 @@ void AnnotationModel::setGroupName(int iGroupIndex,
 
 //=============================================================================================================
 
-const QString& AnnotationModel::getGroupName(int iGroupIndex)
+QString AnnotationModel::getGroupName(int iGroupIndex)
```
LorenzE, line 867: Do we still need this function or is it obsolete due to the new `getGroupNameFromList`?

gabrielbmotta: Currently we are not using it, but they get data in a different way. The old function, getGroupName, gets the name based on the group's map key, while the new one just gets a list of all groups and selects from this list, disregarding the map key. Should we remove it for now and if we need that functionality later we just re-implement it?

LorenzE: ok. so in the second (new) function the function Parameter should maybe be renamed to `iGroupListIndex`? And in the old function maybe we should use `iGroupMapKey`?

PR #670, opened 2020-07-23
[ENH, FIX] Add custom file name to storeFwd() function

RDoerfel:
This PR includes:
- pass the file name to storeFwd() function, so that you can store fwd. solutions with different names. This is primarily an interest in recomputing the forward solution with a new head position.
- revert transposing of the forward solution after storing


libraries/fwd/computeFwd/compute_fwd.cpp
```diff
@@ -2465,7 +2470,7 @@ void ComputeFwd::updateHeadPos(FiffCoordTransOld* transDevHeadOld)
 
 //=========================================================================================================
 
-void ComputeFwd::storeFwd()
+void ComputeFwd::storeFwd(QString sSolName)
```
LorenzE, line 2473: use `const QString& sSolName` instead

libraries/fwd/computeFwd/compute_fwd.h
```diff
@@ -132,8 +132,10 @@ class FWDSHARED_EXPORT ComputeFwd
     //=========================================================================================================
     /**
      * store Forward solution
+     * @param [in] sSolName        The file name to store the currnt forward solution
```
LorenzE, line 135: document the default behavior

PR #688, opened 2020-09-03
[FIX] eegosports impedance check

johaenns:
Moved impedance check and project setup to eegosportssetupwidget, so that they are available before measurement start (as necessary).


applications/mne_scan/plugins/eegosports/eegosports.cpp
```diff
@@ -148,7 +137,7 @@ void EEGoSports::init()
     m_bWriteDriverDebugToFile = false;
     m_bCheckImpedances = false;
 
-    m_sElcFilePath = settings.value(QString("EEGOSPORTS/elcFilePath"), QString(QCoreApplication::applicationDirPath() + "/resources/3DLayouts/standard_waveguard64_duke.elc")).toString();
+    m_sElcFilePath = settings.value(QString("EEGOSPORTS/elcFilePath"), QString(QCoreApplication::applicationDirPath() + "../resources/general/3DLayouts/standard_waveguard64_duke.elc")).toString();
```
LorenzE, line 140: the resources folder should reside in the applications path. If not you have maybe forgot to add the to the plugins resources in the .pro file? You can take a look at how we deploy resources [here](https://mne-cpp.github.io/pages/development/ci_deployment.html#resource-handling).

LorenzE (commented):
How about we create an additional button to start an impedance measurement? The current workflow (by pressing the Play button in the upper left corner) will start the threads of all plugins included in the pipeline. Maybe a better workflow would be:

1) Open the eego plugin
2) Open the impedance view and press the (new) start impedance measurement button. This should only start the ego's thread and producer in impedance mode.
3) Let the user manually stop the impedance measurement session
4) If the impedance session was closed successfully the user can now press the pipeline play/start button

What do you think @johaenns ?

applications/mne_scan/plugins/eegosports/eegosports.cpp
johaenns: Ok, I see. It seems though that not all layouts from mne-cpp/resources/3DLayouts are copied to mne-cpp/bin/resources/3DLayouts. If you could point me to where this can be changed, I'll correct this.

johaenns:
> �How about we create an additional button to start an impedance measurement? The current workflow (by pressing the Play button in the upper left corner) will start the threads of all plugins included in the pipeline. Maybe a better workflow would be:
> 
>     1. Open the eego plugin
> 
>     2. Open the impedance view and press the (new) start impedance measurement button. This should only start the ego's thread and producer in impedance mode.
> 
>     3. Let the user manually stop the impedance measurement session
> 
>     4. If the impedance session was closed successfully the user can now press the pipeline play/start button
> 
> 
> What do you think @johaenns ?

I agree, that's exactly what this commit should implement. The eegosportssetupwidget, which you see before starting a measurement, get's two buttons, "Run Impedance Test" and "Set Electrode Layout".

"Run Impedance Test" opens the eegosportsimpedancewidget, in which you can then start an impedance measurement.

"Set Electrode Layout" allows to set the correct electrode layout before starting the measurement, because once the measurement started the layout information are already set in the fiffinfo, so that it's too late to change it.

applications/mne_scan/plugins/eegosports/eegosports.cpp
LorenzE: This can be changed/adde [here](https://github.com/johaenns/mne-cpp/blob/aec9d068776269da44acdc13fd257b64f6aae444/applications/mne_scan/plugins/eegosports/eegosports.pro#L110)

LorenzE:
> > �How about we create an additional button to start an impedance measurement? The current workflow (by pressing the Play button in the upper left corner) will start the threads of all plugins included in the pipeline. Maybe a better workflow would be:
> > ```
> > 1. Open the eego plugin
> > 
> > 2. Open the impedance view and press the (new) start impedance measurement button. This should only start the ego's thread and producer in impedance mode.
> > 
> > 3. Let the user manually stop the impedance measurement session
> > 
> > 4. If the impedance session was closed successfully the user can now press the pipeline play/start button
> > ```
> > 
> > 
> > What do you think @johaenns ?
> 
> I agree, that's exactly what this commit should implement. The eegosportssetupwidget, which you see before starting a measurement, get's two buttons, "Run Impedance Test" and "Set Electrode Layout".
> 
> "Run Impedance Test" opens the eegosportsimpedancewidget, in which you can then start an impedance measurement.
> 
> "Set Electrode Layout" allows to set the correct electrode layout before starting the measurement, because once the measurement started the layout information are already set in the fiffinfo, so that it's too late to change it.

Ah yes. I forgot that this behavior was already implemented in the impedance widget.

applications/mne_scan/plugins/eegosports/eegosports.cpp
johaenns: Ah, I see. Overlooked that in the document you linked. Is the idea to only copy the one electrode layout that is hard-coded in the plugin to ./bin/resources/ or should also some alternatives that are given in ./resources/ be copied?

LorenzE: Is there a way for the user to change the electrode Lay-out during runtime? If yes then we should add all Lay-out files to be copied.

applications/mne_scan/plugins/eegosports/eegosports.cpp
```diff
@@ -137,7 +137,7 @@ void EEGoSports::init()
     m_bWriteDriverDebugToFile = false;
     m_bCheckImpedances = false;
 
-    m_sElcFilePath = settings.value(QString("EEGOSPORTS/elcFilePath"), QString(QCoreApplication::applicationDirPath() + "../resources/general/3DLayouts/standard_waveguard64_duke.elc")).toString();
+    m_sElcFilePath = settings.value(QString("EEGOSPORTS/elcFilePath"), QString(QCoreApplication::applicationDirPath() + "./resources/general/3DLayouts/standard_waveguard66.elc")).toString();
```
LorenzE, line 140: `./` is not needed since the absolute application path `QCoreApplication::applicationDirPath()` is prepended.

LorenzE:
Thx! We will have to wait until #691 is merged. The CI is currently broken.

LorenzE (approved):

PR #702, opened 2020-09-18
[ENH] MNE-Lib: Implement triangle picking 

RDoerfel:
This PR implements triangle picking of custom mashes, such as the Bem surface.

Therefore following member functions and variables were implemented into DISP3DLIB::View3D.

Public:
- `void activatePicker(const bool bActivatePicker);` activates/deactivates the Picker. Picking can need quite some performance if a lot of Triangles/Vertices/etc. are involved. Therefore, it is helpful only to activate it if necessary.
 

Protected:
- `void initObjectPicking();` - initialize the QObjectPicker
- `void handlePickerPress(Qt3DRender::QPickEvent *qPickEvent);` - called whenever a picking event occure. Emits an signal if the left mouse button is the origin.

Signal:
- `void pickEventOccured(Qt3DRender::QPickEvent *qPickEvent);` Emitten, when picking event occured

### DIFF
--- a/libraries/disp3D/engine/view/view3D.cpp
+++ b/libraries/disp3D/engine/view/view3D.cpp
@@ -66,6 +66,12 @@
 #include <Qt3DExtras/QSphereMesh>
 #include <Qt3DRender/QRenderSettings>
 
+#include <QObjectPicker>
+#include <QPickingSettings>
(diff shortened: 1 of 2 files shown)


LorenzE (approved):
Excellent work @RDoerfel

PR #706, opened 2020-10-07
[FIX] Adjust scaling on Channel Selection

gabrielbmotta:
- Remove max size on widget items, change size preference.


LorenzE:
The current changes would just scale the group boxes in width to fill the residual space between the 2D layout scene and the GUI elements on the right. I think it would be better if the 2D layout scene always occupies the available space.

LorenzE (approved):
Thx @gabrielbmotta

PR #707, opened 2020-10-08
WIP [ENH,MAINT] MNE-BIDS in Library and MNE Analyze

gabrielbmotta:
- AnalyzeDataModel, DataManagerControlView are now BidsViewModel, BidsView.
- Reworked hierarchy of stored items in model to fit bids. Items now hold their own place in structure.
- Added session, and data type folders to view.
- Move sessions between subjects, move data between sessions.
- Loading .eve files
- Loading -ave.fif files
- Add item for annotation model in bidsviewmodel
- reworked instantiation and handling of annotationmodel (no longer crated in fiffrawviewmodel)


LorenzE (changes requested):
first round of review comments. thx @gabrielbmotta

libraries/mne/mne.cpp
```diff
@@ -274,3 +276,23 @@ void MNE::setup_compensators(FiffRawData& raw,
         }
     }
 }
+
+//=============================================================================================================
+
+bool MNE::read_events(QIODevice &p_IODevice,
```
LorenzE, line 282: Please rename this function to `read_events_from_ascii` and make use of `MatrixXi` instead of a `QList`. The equivalent function which reads from .fif file should be renamed to `read_events_from_fif`. Make use of these two functions in `MNE::read_events(QString t_sEventName, QString t_fileRawName,MatrixXi& events)`

LorenzE (changes requested):
some more comments. Please also update the men-cpp website documentation on the data manager plugin.

LorenzE (approved):

PR #746, opened 2021-01-04
[MAINT] MNE Analyze - filter control update

gabrielbmotta:
FilterDesignView/FilterViewSettings:
- Make simple filter control show changes made in filter designer (and vice versa)
- Make simple filter control show defaults from filter designer when initiated.
- Load available filter design methods from FilterKernel

FilterKernel:
- Add new FilterParameter class to replace enums.
- Add static vectors in for design method and filter types to replace enums and static conversion functions.

ScalingView/FiffRawViewSettings:
- Change default loaded values

ty Juan


LorenzE:
Does this go both ways, e.g., are the simple filter values updated when the values change in the filter designer?

gabrielbmotta:
> Does this go both ways, e.g., are the simple filter values updated when the values change in the filter designer?

This is what the pull request is implementing. Currently any changes to the simple filter view are reflected in the filter designer, but changes to filter designer are not reflected in the filter view. This pull request changes it so that the simple filter now gets the initial values from the designer when the filter designer is initiated (previously the simple and designer had different default values) and any updates to the 'From' and 'To' in either view is reflected in the other.

LorenzE:
Ok nice! Did you check if this also works when saving/loading filter settings when closing/starting MNE Analyze? Is the correct filter created when the values are loaded via QSettings? I vaguely remember that there was an issue when I worked on it.

LorenzE (dismissed):
Thx @gabrielbmotta

LorenzE:
+1 from my end

gabrielbmotta:
> Ok nice! Did you check if this also works when saving/loading filter settings when closing/starting MNE Analyze? Is the correct filter created when the values are loaded via QSettings? I vaguely remember that there was an issue when I worked on it.

This seems to be working on closing/starting, but I'm running into issues when loading/saving filters in the designer (mne analyze crashes). I'm checking upstream/master and it seems the error is present there as well.

LorenzE:
I guess this is a new issue. Can you create a new issue please?

gabrielbmotta:
> I guess this is a new issue. Can you create a new issue please?

I think i found a fix

juangpc:
Gabriel, is this stable now on your side?

gabrielbmotta:
> Gabriel, is this stable now on your side?

Yes. The filter GUI updating on both ends, loading/saving, and the new changes we made to the FilterKernel class are running on my end.

juangpc (approved):

PR #751, opened 2021-01-15
[BUG] Orbit camera instead of rotating objects in 3D view

alexrockhill:
Sorry this took so long to have a decent enough start to commit @LorenzE! I think I have the lay of the land and can work on a solution. I'm trying to implement it like here https://doc.qt.io/qt-5/qt3d-simple-cpp-example.html (I tried first with `QCamera::panAboutViewCenter` but I guess you can't animate functions only properties?). It's still crashing but I was going to work on it for another hour or so and then maybe ask for help if it's still breaking.

### DIFF
--- a/libraries/disp3D/engine/view/orbitalcameracontroller.cpp
+++ b/libraries/disp3D/engine/view/orbitalcameracontroller.cpp
@@ -65,7 +65,7 @@ using namespace DISP3DLIB;
 
 OrbitalCameraController::OrbitalCameraController(Qt3DCore::QNode *pParent)
     :QAbstractCameraController(pParent)
-    , m_angle(0.0f)
+    , m_rotating(0.0f)
 {
     initController();
 }
@@ -141,22 +141,16 @@ void OrbitalCameraController::initController()
 
 //=============================================================================================================
 
-void OrbitalCameraController::setAngle(float angle)
+void OrbitalCameraController::setRotating(int count)
 {
     Qt3DRender::QCamera *pCamera = this->camera();
 
-    if (!qFuzzyCompare(angle, m_angle)) {
-        m_angle = angle;
-        pCamera->setPosition(QVector3D(0.0f, -0.4f, -0.25f));
-        pCamera->setViewCenter(QVector3D(0.0f, 0.0f, 0.0f));
-        pCamera->setUpVector(QVector3D(0.0f, 1.0f, 0.0f));
-        pCamera->tiltAboutViewCenter(180);
-        QQuaternion quat = QQuaternion::QQuaternion::fromEulerAngles(0,0,angle);
-        pCamera->rotateAboutViewCenter(quat);
-    }
+    m_rotating = count;
+    QQuaternion quat = QQuaternion::QQuaternion::fromEulerAngles(0,0,0.5);
+    pCamera->rotateAboutViewCenter(quat);
 }
 
-float OrbitalCameraController::angle() const
+int OrbitalCameraController::rotating() const
 {
-    return m_angle;
+    return m_rotating;
 }
--- a/libraries/disp3D/engine/view/orbitalcameracontroller.cpp
+++ b/libraries/disp3D/engine/view/orbitalcameracontroller.cpp
@@ -146,7 +146,7 @@ void OrbitalCameraController::setRotating(int count)
     Qt3DRender::QCamera *pCamera = this->camera();
 
     m_rotating = count;
-    QQuaternion quat = QQuaternion::QQuaternion::fromEulerAngles(0,0,0.5);
+    QQuaternion quat = QQuaternion::QQuaternion::fromEulerAngles(0,0,m_fAutoRotationSpeed);
     pCamera->rotateAboutViewCenter(quat);
 }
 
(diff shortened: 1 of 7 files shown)


alexrockhill:
Hmmm it's crashing without an error so I'll have to figure out the debugger...

alexrockhill:
Ok, ready for review!

LorenzE:
I would suggest making the OrbitalCameraController’s property to take in the angle to rotate in degrees instead of a plain int. This way the rotate functions are a bit more descriptive and more functional. You would also be able to remove the member variable `m_rotating`.

alexrockhill:
Ok so I checked and you get a warning if the `QProperty` doesn't have a READ function and you get an error if you try not to pass a value to `QPropertyAnimation` so I think this is the best that can be done. Alternatively, we could return `null` for reading in the property `rotating` but that seems more confusing than the way it is now.

libraries/disp3D/engine/view/orbitalcameracontroller.cpp
```diff
@@ -138,3 +140,17 @@ void OrbitalCameraController::initController()
...
+
+void OrbitalCameraController::setRotating(int count)
+{
+    Qt3DRender::QCamera *pCamera = this->camera();
+
+    m_rotating = count;
+    QQuaternion quat = QQuaternion::QQuaternion::fromEulerAngles(0,0,m_fAutoRotationSpeed);
+    pCamera->rotateAboutViewCenter(quat);
+}
+
```
LorenzE, line 152: Please add spacer

libraries/disp3D/engine/view/orbitalcameracontroller.h
```diff
@@ -107,6 +115,13 @@ class DISP3DSHARED_EXPORT OrbitalCameraController : public Qt3DExtras::QAbstract
      */
     void invertCameraRotation(bool newStatusFlag);
 
+    //=========================================================================================================
+    /**
+     * Sets the angle of the camera for rotating around
+     */
+    void setRotating(int count);     /* The counter for how long the rotation has been happening. */
+    int rotating() const;
```
LorenzE, line 123: please document each function and their functions values (input and return)

libraries/disp3D/engine/view/orbitalcameracontroller.h
```diff
@@ -136,6 +151,8 @@ class DISP3DSHARED_EXPORT OrbitalCameraController : public Qt3DExtras::QAbstract
 
     float m_rotationInversFactor = 1.0f;             /**< The factor used to invers the camera rotation. */
```
LorenzE, line 152: Please mind indentation for start of docu /**<

LorenzE, line 152: Although you did not add this line, please rename `m_rotationInversFactor` to `m_fRotationInverseFactor`. the `f` hereby indicates a float value

libraries/disp3D/engine/view/orbitalcameracontroller.h
```diff
@@ -136,6 +151,8 @@ class DISP3DSHARED_EXPORT OrbitalCameraController : public Qt3DExtras::QAbstract
...
     const float m_fZoomInLimit = 0.04f;         /**< The minimum distance of the camera to the the view center. */
+    const float m_fAutoRotationSpeed = 1.0f;         /**< The speed that automatic rotation rotates */
+    int m_rotating;                      /**< How long the camera has been rotating with regards to the view center. */
```
LorenzE, line 155: Please rename to m_iRotating

libraries/disp3D/engine/view/view3D.h
```diff
@@ -122,9 +123,9 @@ class DISP3DSHARED_EXPORT View3D : public Qt3DExtras::Qt3DWindow
 
     //=========================================================================================================
     /**
-     * Starts or stops to rotate all loaded 3D models.
+     * Starts or stops to rotate camera around 3D models.
      */
-    void startStopModelRotation(bool checked);
+    void startStopCameraRotation(bool checked);
```
LorenzE, line 128: rename `checked` to `bChecked`

libraries/disp3D/engine/view/view3D.h
```diff
@@ -51,6 +51,7 @@
 #include <QPointer>
 #include <QObjectPicker>
 #include <Qt3DCore>
+#include "orbitalcameracontroller.h"
```
LorenzE, line 54: move `#include "orbitalcameracontroller.h"` to line 44 right after `#include "../../disp3D_global.h"`

LorenzE (changes requested):
Looks good to me. Just some minor comments

libraries/disp3D/engine/view/orbitalcameracontroller.h
alexrockhill: I forgot that one! I wasn't sure when I was trying things if they would work so I didn't document but I forgot to go back and fix it

alexrockhill:
Ok should be good to go!

LorenzE (approved):
Nice work @alexrockhill

PR #752, opened 2021-01-16
[FIX] Update design filter freq response view

juangpc:
Two problems. When filterdesignview is created the filter freq response graph is also created and hidden. 
When a users clicks the button, the window is unhidden and, at least in windows, you see this:
 - 
Screenshot 2021-01-15 165728

So we: 
 - make updateFilterPlot part of the public api. 
 - disable the scroll bar vert and horiz, because they kind of looked bad.


LorenzE (dismissed):

gabrielbmotta (approved):
👍

gabrielbmotta:
CI failing because of [Qt download issue](https://lists.qt-project.org/pipermail/development/2021-January/040912.html)

LorenzE:
https://github.com/jurplel/install-qt-action/issues/78

PR #756, opened 2021-01-25
[FIX]  MNEAnonymize Use QDate instead of QDateTime when reffering to birthday date.

juangpc:
In fiff files, birthday date is coded in julian calendar (with an integer number).  It is just a date, not a date and time. Thus, I change here the use of QDateTime for the use of QDate for a more correct use. 

I also got to change the use of const references more to gain a bit of simplicity/efficiency. 

Let's see what I break.


LorenzE (approved):

juangpc:
That was fast. Thx

PR #758, opened 2021-01-26
[ENH] Cross platform deployment scripts, accepting linkage option, callable from anywhere and segregating deploy and packing.

juangpc:
During everyday tasks, I find myself using more and more this script. 
Want to make them to be able to be used from anywhere. 
Also, usually, I don't want to have the zip folder created so I've offloaded that specific task to another script. 

@LorenzE @gabrielbmotta , could you take a look, even a more thorough than usual, to [this part](https://github.com/juangpc/mne-cpp/blob/b5eaabf8b1071ed53b7bb5abda0326ce0d4a09f4/.github/workflows/release.yml#L219)?
```
    - name: Deploy binaries
      run: |
        ./tools/deployment/deploy_win_static.bat
    - name: Generate zip file from bin folder contents
      run: |
        ./tools/deployment/save_bin_folder_content_as_zip_file.bat   
```

and [this](https://github.com/juangpc/mne-cpp/blob/b5eaabf8b1071ed53b7bb5abda0326ce0d4a09f4/.github/workflows/release.yml#L416): 
```
    - name: Deploy binaries (Windows)
      run: |
        ./tools/deployment/deploy_win.bat
    - name: Generate zip file from bin folder contents
      run: |
        ./tools/deployment/save_bin_folder_content_as_zip_file.bat
```


LorenzE:
Very nice idea! Some thoughts:

- Rename `save_bin_folder_content_as_zip_file.bat` to `package_win.bat`. 
- Also create `package_win_static.bat`, `package_linux`, `package_linux_static`, `package_macos`, `package_macos_static`. We would need the static package scripts because the names of the zipped files differ from the dynamic ones.
- Move the removing of files from the deploy scripts to the package scripts. removing of files is only done because we do not want to package/zip them. Moving the remove commands from deploy to package would also make sense because the developer running, e.g., the win_deploy.bat most likely does not want to remove the sample data.
- The package scripts should be included in the pullrequest.yml as well.
- I do not understand why the CI is failing at the `Remove-Item` call. Have you tried to run the `Remove-Item` command with `%scriptPath%` locally?

And now a beer. Prost! 🍻

juangpc:
Hi, 
> I do not understand why the CI is failing at the Remove-Item call. Have you tried to run the Remove-Item command with %scriptPath% locally?

It is working now. Using back and normal slash at the same time seemed to confuse the thing.

> Rename save_bin_folder_content_as_zip_file.bat to package_win.bat.

I understand that someone might not like the name I chose. But, if possible, I would like it to be descriptive. ```package_win.bat`` seems too cryptic for me. How about ```pack_bin_folder.bat```

>Also create package_win_static.bat, package_linux, package_linux_static, package_macos, package_macos_static. We would need the static package scripts because the names of the zipped files differ from the dynamic ones.
Move the removing of files from the deploy scripts to the package scripts. removing of files is only done because we do not want to package/zip them. Moving the remove commands from deploy to package would also make sense because the developer running, e.g., the win_deploy.bat most likely does not want to remove the sample data.
The package scripts should be included in the pullrequest.yml as well.

I disagree with this. As with the name of the scrip, I think we need to strive for ease of use and simplicity. So I went and create a new script which is valid in linux (bash and sh shells, have not tried zsh, but probably will work), windows and mac (have not tried yet). It also accepts a parameter to specify the linkage option.

So instead of having 12 scripts/files we have 2. One for deploy and one for pack. 

Whenever I get my hands on my mac I'll try the mac version.
The deployment scripts still have to be ported to this cross platform way.

juangpc:
If someone wants to debug and finish the macos section of the script. Feel free.

LorenzE:
`pack_bin_folder.bat` works for me, too. Having fewer scripts taking an input parameter would be neat! Although the inside of the scripts with input parameters seem to be more complex than compared to having script names such as `package_win.bat`. But still I would prefer having fewer scripts with a more complex heuristic inside.

juangpc:
> `pack_bin_folder.bat` works for me, too. Having scripts taking an input parameter would be neat! Although the inside of the scripts with input parameters seem to be more complex than compared to having script names such as `package_win.bat`. But still I would prefer having fewer scripts with a more complex heuristic inside over 12 individual scripts. Do you have a Mac at your disposal to debug? Once you have a working prototype I would be happy to give it a try on my Mac book, too.

Great. I'll use `pack_bin_folder.bat`. 
Yes I do. I'll try sometime this week. Apart from testing locally, it all depends on the default shell used in the github vm. Hope it works. It took a while to figure it out.

juangpc:
I've a doubt. 
Shouldn't this command be packing `bin/`, instead of `mne-cpp/` folder? Maybe I'm missing something?

LorenzE:
I think this should be fine. In case of linux we not only copy the bin folder but also the lib folder. So you end up with two folders bin and lib in the tar file.

juangpc:
> I think this should be fine. In case of linux we not only copy the bin folder but also the lib folder. So you end up with two folders bin and lib in the tar file.

Thank you. 

Another (related) doubt. Folder deletions now differ between dynamic and static deploy scripts in linux.  
Since we're having only one script for deployment of both versions (dynamic and static) which option should I choose?

https://github.com/mne-tools/mne-cpp/blob/c8e0124fd7630b5daff6477309a819d8b9dbfa90/tools/deployment/deploy_linux_static#L32

juangpc:
Doubts: 
 - Deployment in linux(!), depends on a linuxdeployqt-continuous-x86_64.AppImage application. If it works that's fine. Just checking.
 - Don't understand the difference between linux static and dynamic, as it is right now.

Comments: 
- For convenience during this revision, I've kept the old scripts as they are right now (8e2dff0e0ad81c529d654eaaff2f173d49565aba) in master branch . Right before merging, these should be deleted and only kept `deploy.bat` and `pack_bin_folder.bat`.
- `release.yml` has been updated. If it doesn't work for some reason I'll try invoking bash/cmd explicitly.
- I'm doing some changes to the deploy script in linux. Because as it is right now it diverges from windows and macos versions, where the deployment changes are poured into your working mne-cpp/bin folder. This has the advantage (as pointed by @LorenzE) that if you want to double click on an app (i.e. mne_analyze) without using qtCreator, after you deploy, you can resolve all dependencies. 
Right now, it seems to me, that is not the case in linux.  It seems like linux script creates a separate folder for deployment. I think it is useful to run directly the app without use of qtCreator, even in a dev environment. Since all changes take place in `bin/` and `lib/` folders, .gitignore deals with whatever dll we add without interaction with the repo.
- The differences between dynamic and static seem to be more in the actual deployment of qt itself (`release.yml`) than in the actual deployment script we're dealing with here. Is that correct? The main differences  seem to be in the actual name used to save the tar/zip file? If that is the case deploy.bat script might not need an input command at all. Is this correct? If this is correct, then the main changes in the `deploy.bat` script will be to make it (1) cross-platform and (2) callable from anywhere.
- Doing this, makes me realize to what level of detail one needs to dive into to have all this "automatic" infrastructure working.
- I've noticed that right now the ci, has @LorenzE github user for clones and checkouts. To me, this is perfectly fine. Thanks for that.

LorenzE:
> Doubts:
> 
> * Deployment in linux(!), depends on a linuxdeployqt-continuous-x86_64.AppImage application. If it works that's fine. Just checking.
> * Don't understand the difference between linux static and dynamic, as it is right now. I mean, in the deployment scripts.

The only difference is that static linux builds do not support the LSL and Brainflow MNE Scan plugins. Also in static builds we can remove the mne scan and mne analyze plugins folder in the bin folder since the plugins have been statically linked (and thereby) included into the executables.

> 
> Comments:
> 
> * For convenience during this revision, I've kept the old scripts as they are right now ([8e2dff0](https://github.com/mne-tools/mne-cpp/commit/8e2dff0e0ad81c529d654eaaff2f173d49565aba)) in master branch . Right before merging, these should be deleted and only kept `deploy.bat` and `pack_bin_folder.bat`.
> * `release.yml` has been updated. If it doesn't work for some reason I'll try invoking bash/cmd explicitly.
> * I'm doing some changes to the deploy script in linux. Because as it is right now it diverges from windows and macos versions, where the deployment changes are poured into your working mne-cpp/bin folder. This has the advantage (as pointed by @LorenzE) that if you want to double click on an app (i.e. mne_analyze) without using qtCreator, after you deploy, you can resolve all dependencies.

I think under linux it more common to have the binaries separated from .lib and .so files. That is why we chose the /lib and /bin folder. I remember Rob from Martinos had problems with the folder structure. Maybe you can reach out to him?

>   Right now, it seems to me, that is not the case in linux.  It seems like linux script creates a separate folder for deployment. I think it is useful to run directly the app without use of qtCreator, even in a dev environment. Since all changes take place in `bin/` and `lib/` folders, .gitignore deals with whatever dll we add without interaction with the repo.
> * The differences between dynamic and static seem to be more in the actual deployment of qt itself (`release.yml`) than in the actual deployment script we're dealing with here. Is that correct? The main differences  seem to be in the actual name used to save the tar/zip file? If that is the case deploy.bat script might not need an input command at all. Is this correct? If this is correct, then the main changes in the `deploy.bat` script will be to make it (1) cross-platform and (2) callable from anywhere.

Yes , the main difference are the names, but also we can safely delete the plugins when building statically as described above.

> * Doing this, makes me realize to what level of detail one needs to dive into to have all this "automatic" infrastructure working.

The release and pullrequest files are almost equal except for the fact that we use github actions matrix build in the pullrequest file and we do not actually upload anything (dev or release builds) to the github repo.

> * I've noticed that right now the ci, has @LorenzE github user for clones and checkouts. To me, this is perfectly fine. Thanks for that.

juangpc:
Hi, 

Going through all this again makes me once more realize to what level of detail the ci has been tuned. Going through the ci ymls I see how you have also prepared code in case we need to compile Qt locally. Thanks also for that. 

I still have some doubts in relation to the deploy_win_static.bat script. 

The last time we left this, we had agreed to try to compound 12 eventual scripts into 2. However I think now that since we are dealing with input parameters we can then use only one script for all operating systems, both linking options and both packing into a zip the end binaries or not. It is really not that complicated. 

- Each platform (linux/mac/win) will run a specific section of the script. 
- There are two input arguments, one to define the linkage and another to specify whether or not to pack into a zip file.
- By default the script **does not pack**. If you want it to, you **must** use a second argument with value "pack".

```deploy.bat dynamic```  -> deploys the dynamic version.
```deploy.bat static``` -> deploys the static version.
```deploy.bat dynamic pack``` -> deploys the dynamic version in the platform used to call the script and then creates the tar/zip file.
```deploy.bat static pack``` -> deploys the static version in the platform used to call the script and then creates the tar/zip file.

The script is ready ( I think). I would need help to go through the results. The platform independent scripting part is working. The things related to the project are the ones that are not quite there.

In windows I think it is working but I get an error regarding the location of the visual studio compiler installation directory. 
Regardless of this, I'm not sure if I need to go through all the processes in "CreateRelease". A few other jobs "need" that other job. So I'm not sure. A bit of help would be appreciated.

**Comment by juangpc** [comment]
Hi,
This is almost merge ready. 

- The tests for the ci are passed for all platforms static and dynamic.
- There is a small error in the windows platform because the variable VCINSTALLDIR is not set.
This could be in ```C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\``` but I'm not sure.
                             
- It seems like we are using msvc2019 for the downloadable code, so it makes sense to use that same compiler for the development. I'm glad we're using that compiler @gabrielbmotta and i. That is also the recommended version in the development tutorial. Good.
- We were using a command ```remove-item``` inside the windows versions which is a powerShell command, but the ci uses a cmd prompt. In cmd, ```remove-item``` results in an error and therefore I'm not sure for how long we've been distributing the releases with data, libraries, includes, which were not supposed to be there. 
- It seems like we are creating a file ```mne-cpp-doc-qtcreator.qch``` which is 187MB(!)  I understand that is the doxygen data. I don't how hard or difficult it would be to delete that file for previous releases. I guess MS can store it, but if it easy enough I think we should do it.

**Comment by juangpc** [comment]
Thanks for today's hands on: https://github.com/juangpc/mne-cpp/actions/runs/619991755

**Comment by LorenzE** [comment]
> - We were using a command ```remove-item``` inside the windows versions which is a powerShell command, but the ci uses a cmd prompt. In cmd, ```remove-item``` results in an error and therefore I'm not sure for how long we've been distributing the releases with data, libraries, includes, which were not supposed to be there. 

Are you sure the GitHub Windows runner uses. I am pretty sure that they used to use PS. But they might have changed it?
 
> - It seems like we are creating a file ```mne-cpp-doc-qtcreator.qch``` which is 187MB(!)  I understand that is the doxygen data. I don't how hard or difficult it would be to delete that file for previous releases. I guess MS can store it, but if it easy enough I think we should do it. 

Yeah, based on the amount CI runs we trigger that is quite a lot. I would not delete it for previous releases, for the sake of completeness. MS Azure's cloud should be able to handle it ;) Btw (and in case you did not know this already) the qch file can be loaded into the QtCreator and you can press F1 to get documentation about the selected object. The same holds true for Qt functions.

**Comment by juangpc** [comment]
Yes I think it has been cmd all along. The script is called ```.batch```, maybe if we were to call them ```ps1```, but I'm not sure. And it is working just fine now. 

But just for the sake of verifying. This is a random master commit I've selected. https://github.com/mne-tools/mne-cpp/runs/1705631749?check_suite_focus=true
It is only a couple of months old because logs get deleted by github after more than that. If you look at the ```Deploy binaries``` part from ```WinStatic``` you'll see that it wasn't recognizing the comment ```#``` sign, and neither the ```Remove-item``` command. The static script bas basically jumping down to build the zip folder: 

image

**Comment by juangpc** [comment]
@LorenzE @gabrielbmotta how should we proceed from this point? Is there anything else we should try test?

In the ```pullrequest.yml``` action we no longer have to call a separate script for each matrix. Good. However, do we need to pack in that case? Aren't we only checking for compilation? I get that the info will get destroyed either way, but at least for the sake of environmental purposes... Should we? Check this commit. ff9ef68f494238c1bb8d6cc6ce7972e9785006c6
In other words, could we use ```deploy.bat dynamic```  --> without ```pack``` as second arg?

The same goes for ```testci.yml``` action.

If anybody could go through the documentation updates (6c6f95e0b58db3dbfca86d92246fbafa9ad75b7e) That also would be much appreciated.

File: doc/gh-pages/pages/development/ci_deployment.md [hunk 1]
```diff
@@ -19,24 +19,32 @@ MNE-CPP depends on [Qt](https://www.qt.io/){:target="_blank" rel="noopener"} and
...
\ No newline at end of file
+The information above corresponds to deploying the dynamically build version of MNE-CPP. In case of static deployment the process differs a bit. See the [deploy.bat](https://github.com/mne-tools/mne-cpp/blob/master/tools/deployment/deploy.bat){:target="_blank" rel="noopener"} file.
+
+## Deployment script `deploy.bat`
+The script used by MNE-CPP to perform the deployment in the three platforms to which MNE-CPP is compiled is the same `deploy.bat` file, which is inside `tools/deployment` folder. 
+The way this is achieved is by writing `deploy.bat` in a way in which different operating systems running the same script will actually execute different parts of it. If you take a look at its contents it is easy to identify the parts executed by each operating system because they are marked with comments. 
+
+The script is also the same indendently of the type of linkage needed during the deployment. It has a first input argument which accepts either ```dynamic``` or ```static``` in order to decide which type of linkage to perform.
```
gabrielbmotta, line 48: > indendently

typo

gabrielbmotta (changes requested):
Just one small nitpick

juangpc:
As far as I understand, this is ready for merge.

gabrielbmotta (approved):

LorenzE:
Great job! This is really cool.

PR #759, opened 2021-01-29
WIP [ENH] Generate PDFs from .md files

gabrielbmotta:
Note: After trying to do it myself with pandoc, better results were achieved with this gitub action. Still, images do not seem to work, and there are a few issues with the pdfs. 

- Convert .md files into pdfs

To do:
- figure out which files we want, when we want to save, where we want to save them, and whether the current state of these pdfs is acceptable. [Preview here.](https://github.com/gabrielbmotta/mne-cpp/actions/runs/521329645)

### DIFF
--- a/.github/workflows/pdfconversion.yml
+++ b/.github/workflows/pdfconversion.yml
@@ -11,9 +11,9 @@ jobs:
     steps:
     - name: Install components
       run: |
-        apt-get -y install pandoc
-        apt-get -y install texlive-latex-base
-        apt-get -y install texlive-fonts-recommended
+        sudo apt-get -y install pandoc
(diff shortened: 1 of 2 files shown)


juangpc:
Hi @gabrielbmotta , I think this is an advancement however: 
- to have a set of unordered pdf documents is a step forward. 
- could it be that the images are not being included because of the path changes from the when it is in the repo and doc branch, to when the wiki is transformed copied to master branch in mne-cpp and then transformed as html and served? 
- I'm very surprised this is not easy to achieve. Your solution  would we just need to use a cl linux app to concatenate pdf docs into a single one? That is a great solution for now. But what about the links between sections. I'm surprised this hasn't been figured out already.

PR #763, opened 2021-02-19
[FIX] MNE Scan - writing to file calibration values

gabrielbmotta:
- Fix writing to file with wrong values by not accounting for calibration (calling correct write_raw_buffer)

### DIFF
--- a/applications/mne_scan/plugins/writetofile/writetofile.cpp
+++ b/applications/mne_scan/plugins/writetofile/writetofile.cpp
@@ -289,7 +289,7 @@ void WriteToFile::run()
                     }
 
                     if(m_pOutfid) {
-                        m_pOutfid->write_raw_buffer(matData);
+                        m_pOutfid->write_raw_buffer(matData, m_mCals);
                     }
                 } else {
                     size = 0;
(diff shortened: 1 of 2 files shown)


LorenzE:
Please double check if this is really a bug. I think the actual data in fiff is stored uncalibrated (raw). The calibration is stored alongside in the fiff info and applied again when reading from the fiff file again. When I doubt please ask Matti about this.

gabrielbmotta:
> Please double check if this is really a bug. I think the actual data in fiff is stored uncalibrated (raw). The calibration is stored alongside in the fiff info and applied again when reading from the fiff file again. When in doubt please ask Matti about this.

In the version in master, we are currently saving the calibration as something in the order of 10^-9 and the raw data as-is. Whenever we load the data( in mne analyze or rt_server), we get values in the order of 10-21 - 10-23 ish.

Juan and I have been hunting this issue for a while now. 😅 We narrowed it down to the saving by going thorough multiple fiff files in a hex viewer, and comparing the calibration values and data values. The files produced by the plugin were not conforming to what we were expecting based on the fiff pdf.

With these changes (using the method with cal as an input), we now apply the inverse calibration before saving, and when reading we can read the correct data. Another solution would be to save the data values as is and instead set cal to 1 when saving the channel parameters. We went with this one since it had an already existing method in the library (and is what mne analyze is using to save to file)

We have an in person meeting with Matti on Monday, we'll ask to make sure.
Sorry for the poor explanation, this is hard to explain over text.

juangpc (dismissed):
After all comments, I think we all agree this is ready for release version.

juangpc (approved):

PR #768, opened 2021-03-02
[ENH] Use explicit copy elision in plugin::clone()

juangpc:
Hi, 

Some weeks ago I proposed this changed in another unrelated pr. Here is now a specific one for discussing this change. 

Going through mne_analyze plugin docs, I saw the code snippet for [the clone method](https://mne-cpp.github.io/pages/development/analyze_plugin.html). I think that this is OK, however it is improvable, due to the fact that: 

 - Since it is a part of the code that is actually described in the documentation we should have special care. 
 - The code is not wrong, but I don't think it is correct either. I haven't seen the asm but I'm quite certain that the compiler will get that code referred to the clone method and disregard it's implementation. 
 - The change I propose makes explicit what the reader should know and understand as part of the c++ standard because it is there and it is great. What we see in the code is an x-value, which will probably not be implemented at all in the compiled code.
 - Even if we want to go with other options, like using create static method from the QSharedPointer class, I still don't think that is 100% correct. What the code describes is to create a local variable, then copy it as a smart shared pointer and let it get deleted by running out of scope. Supposedly the counter in the smart pointer will be decreased, but that is not what the standard says and that is not what will happen.
 
So these are the reasons for the change. What do you guys think?

### DIFF
--- a/applications/mne_analyze/plugins/averaging/averaging.cpp
+++ b/applications/mne_analyze/plugins/averaging/averaging.cpp
@@ -115,8 +115,7 @@ Averaging::~Averaging()
 
 QSharedPointer<AbstractPlugin> Averaging::clone() const
 {
-    QSharedPointer<Averaging> pAveragingClone(new Averaging);
-    return pAveragingClone;
+    return QSharedPointer<AbstractPlugin> (new Averaging);
 }
 
 //=============================================================================================================
--- a/doc/gh-pages/pages/development/analyze_plugin.md
+++ b/doc/gh-pages/pages/development/analyze_plugin.md
@@ -25,8 +25,7 @@ AbstractPlugin has a number of pure virtual functions that need to be defined by
 Returns an instance of the plugin. This is not a copy. Most of the existing plugins do something like this:
 
 ```
-QSharedPointer<myNewPlugin> pMyNewPluginClone(new myNewPluginClone);
-return pMyNewPluginClone;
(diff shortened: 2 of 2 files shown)


juangpc:
I'll try to explain myself better:

This is the code you find in a plugin, and in the documentation: 
```c++
QSharedPointer<myNewPlugin> pMyNewPluginClone(new myNewPluginClone);
return pMyNewPluginClone;
```
The problem I find with this is that you're generating a shared pointer which will be destroyed immediately, decreasing its counter, etc... but only after the caller of the clone method will create a copy of that pointer, increasing the counter just in time,. This is OK, but not the best practice. However that is not what troubles me. What I really think is that the compiler will not end up doing that at all because of return value optimization. The object x-value will be created directly and only once in the callers context, saving a copy. 
This is the option I propose to use. This makes it clear how c++ works. 

```c++
return QSharedPointer<AbstractPlugin> (new Averaging);
```

Probably this would even work. Haven't tested it, but it will work, perhaps it needs some void* casting. However this is probably too idiomatic. 

 ```c++
return new Averaging;
```

LorenzE:
I agree returning `QSharedPointer<AbstractPlugin> (new Averaging);` should be fine. I do not think the QSharedPointer constructor is defined with the explicit keyword so you `return new Averaging;` should work?! Interesting! Let me know what you find out :)

gabrielbmotta (approved):

PR #770, opened 2021-03-08
[ENH] Update brainflow version

Andrey1994:


LorenzE:
The only problematic thing I could find is here https://github.com/mne-tools/mne-cpp/runs/2077918755#step:9:265

- Virtual memory exhausted? Did you ever encounter similar out of memory problems on a local build @Andrey1994?

Andrey1994:
No, I see it for the first time

Andrey1994:
Maybe you can try without -j, in Brainflow ci it compiles fine, but I dont use ubuntu 16 images. And I guess mne-cpp builds more code than here

Andrey1994:
@LorenzE I removed -j and seems like it works fine now. There is one cancelled job and it doesnt look relevant to this PR

LorenzE:
> @LorenzE I removed -j and seems like it works fine now. There is one cancelled job and it doesnt look relevant to this PR

I restarted the checks. I don not think removing `-j` did the trick: https://github.com/mne-tools/mne-cpp/pull/770/checks?check_run_id=2212966517#step:9:291

I do not understand why under ubuntu 16.04 and Qt 5.15.2 the Brainflow build takes so long that the runner runs out of memory. For ubuntu 16.04 and Qt 5.10.2 it works just fine.

juangpc:
Hi,

the problem might be that in ```src/utils/serial_ioctl.cpp``` you are using ```__has_include_``` which is [c++17](https://en.cppreference.com/w/cpp/preprocessor/include). However your makefile specifies 11?

Although I'm not so sure about this, On top of using a cpp17 feature, it seems that gcc version installed in ubuntu-16.04 is 4.5 and it was in version 5 when ```__has_include_``` was made compatible. [see](https://gcc.gnu.org/gcc-5/changes.html)

I've tested: 
 - with gcc 4.8 (4.8.5 20150623 (Red Hat 4.8.5-44)) 
 - commenting the lines and including directly the ioctls and termbits files.

And it seems to work.

```c++ 
// linux, also need to check that system headers installed
#elif defined(__linux__) && !defined(__ANDROID__)
//#if __has_include(<sys/ioctl.h>) && __has_include(</usr/include/asm/ioctls.h>) && __has_include(</usr/include/asm/termbits.h>)
#include <sys/ioctl.h>

#include </usr/include/asm/ioctls.h>
#include </usr/include/asm/termbits.h>
//#else
//#define NO_IOCTL_HEADERS
//#endif
```

Andrey1994:
Hi, thanks for looking at this.

<< the problem might be that in src/utils/serial_ioctl.cpp you are using __has_include_ which is c++17. However your makefile specifies 11?

You are right, from the beginning brainflow used C++11, __has_include was added a little later. In C++ standards its indeed was added in C++17, but I use it only for linux and only for gcc. In GCC __has_include was added in gcc 5.0, gcc 5.0 doesnt support all features from C++17(and more likely even C++ 14) but has this one which I use. Its why I didnt bump C++ standard in CMakeLists

<<  it seems that gcc version installed in ubuntu-16.04 is 4.5

I thought in ubuntu 16 there is gcc 5.0+... I downloaded raw logs from failed pipeline and there is gcc 5.5(https://pipelines.actions.githubusercontent.com/EhaHx68s3U3H7RvAT1z0aoJae5ZXlXKZmKzTy2OVXDNOcYhmAx/_apis/pipelines/1/runs/1692/signedlogcontent/7?urlExpires=2021-04-09T14%3A43%3A53.7789395Z&urlSigningMethod=HMACV1&urlSignature=o7mFvsgHoeQARZ6ovp4LSSRCwKDbT6sufPzxTxk%2BOd8%3D)

```
The C compiler identification is GNU 5.5.0
2021-03-31T14:46:19.1978723Z -- The CXX compiler identification is GNU 5.5.0
2021-03-31T14:46:19.2102975Z -- Detecting C compiler ABI info
2021-03-31T14:46:19.2678844Z -- Detecting C compiler ABI info - done
2021-03-31T14:46:19.2794554Z -- Check for working C compiler: /usr/bin/cc - skipped
2021-03-31T14:46:19.2799365Z -- Detecting C compile features
2021-03-31T14:46:19.2810855Z -- Detecting C compile features - done
2021-03-31T14:46:19.2862812Z -- Detecting CXX compiler ABI info
2021-03-31T14:46:19.3603892Z -- Detecting CXX compiler ABI info - done
2021-03-31T14:46:19.3724213Z -- Check for working CXX compiler: /usr/bin/c++ - skipped
2021-03-31T14:46:19.3725581Z -- Detecting CXX compile features
2021-03-31T14:46:19.3732449Z -- Detecting CXX compile features - done
2021-03-31T14:46:19.3746699Z 64 bits compiler detected
```

Also, in case of using unsupported method error message should be different I think

Andrey1994:
In this update I added eigen library as a dependency and I suppose now gcc needs a little more memory to compile it.

And probably on these VMs there is not enough RAM, I will try to check and maybe optimize it.
Another potential fix - build it using clang or maybe try ninja instead make.
Probably its possible to enable swap inside CI runner

Andrey1994:
I fixed it, I forgot to remove `-j` in one job

LorenzE:
So it took significantly more memory to build with multiple threads enabled?

Andrey1994:
<< So it took significantly more memory to build with multiple threads enabled?

Exactly, when you compile with multiple threads it consumes much more RAM. I knew it but forgot to remove -j in one job first. Its common, but maybe here its even a little worse since I copy binary files inside CMake(https://gitlab.kitware.com/cmake/community/-/wikis/contrib/macros/CopyIfDifferent)

Andrey1994:
I will also add this check https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html for __has_include, it will restore support for gcc 4.x

LorenzE (approved):
Thx @Andrey1994

PR #772, opened 2021-03-11
[DOC, WIP] Uplift doxygen documentation standards

juangpc:
Hi, while solving other prs I figured it was worth trying to see how much could the documentation be improved through regular expressions. 

I think in its majority this is an improvement. However I see a few problems (as others might probably see too). The volume of files is uncomfortable, and it is difficult to assess how/if other errors are introduced this way. Another problem is that I've tried to be sure that when I add the ```in``` tag to an argument it is not a ```in,out```. But I haven't able to thoroughly consider every single input argument. However, it is true that before there was no tag at all, so in any case this would mean the added new tag is (only) half-wrong. But yes, it is uncomfortable. 

Let's comment it. I'm not sure this is at all a good idea myself and it has not been a whole lot of effort. So we can always drop it.


LorenzE:
@juangpc This is a good initiative! I agree adding [in] where not present is better than none. Please let me know once this PR is in a state where you would like to merge.

juangpc:
> @juangpc This is a good initiative! I agree adding [in] where not present is better than none. Please let me know once this PR is in a state where you would like to merge.

This is a good initiative by @gabrielbmotta!

gabrielbmotta (approved):

LorenzE:
Why was this closed?

juangpc:
🙃

gabrielbmotta (approved):

PR #773, opened 2021-03-20
[WIP,FIX] BabyMeg HPI Fitting

RDoerfel:
This PR tries to fix the ongoing problems with the BabyMeg HPI fitting. After quite a lot of debugging and desperate nights, I was able to fix some stuff and get at least stable fits, also with a 'high' error. The fits are quite stable around 10 mm displacement error. 

Nevertheless, this is still more than with the old 42ce89b version before any changes were introduced. Besides a lot of other things, I also ended up copying the entire old HPI class and embed it into the current version. The results were as bad as with the most recent version. From that, I conclude that the problem was introduced with the new HPI plugin and I would suggest checking what is the difference between the way the HPI plugin and former `rthpis` and `HPIView`.

This brings us to what is yet in this PR:

HPI Fitting Class:

- use absolute value to determine channel with maximum amplitude

HPI Plugin:

- projectors and compensators were never updated and applied not entirely correct. The way now is more similar to how it was done before. 

That's it so far. I'm still not sure what will fix it entirely. It must be something that only affects the BabyMeg, since the VectorView fits are running like a charm. Also, the good results with the current versions are only obtained using the advanced model, and not the fast version.

### DIFF
--- a/applications/mne_scan/plugins/hpi/hpi.cpp
+++ b/applications/mne_scan/plugins/hpi/hpi.cpp
@@ -191,7 +191,13 @@ void Hpi::update(SCMEASLIB::Measurement::SPtr pMeasurement)
         // Check if data is present
         if(pRTMSA->getMultiSampleArray().size() > 0) {
             //If bad channels changed, recalcluate projectors
-            updateProjections();
+            if(m_iNumberBadChannels != m_pFiffInfo->bads.size()
+                || m_matCompProjectors.rows() == 0
+                || m_matCompProjectors.cols() == 0) {
+                updateProjections();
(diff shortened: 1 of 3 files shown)


juangpc:
This is great, thank you. Even if the problem is (partly) there, to narrow down the probable cause is extremely helpful.

juangpc (approved):

RDoerfel:
No problem. I will try to work further on it in the next one or two weeks if it does not bother to keep the PR open a bit longer :)

juangpc:
Absolutely not. As shared before, your contribution could become part of the essential toolset of every meg lab. This is great. Take all the time you need.

LorenzE:
Was this code ever tried on the Babymeg?

PR #776, opened 2021-03-28
[ENH] Add new CI trigger

LorenzE:
Add new trigger to be able to start Github action workflows manually. This way we do not have to rerun the whole test with all workflows.


LorenzE:
Closing this PR. This is not what I was looking for after all -> https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

PR #779, opened 2021-04-07
[ENH] EventManager sintactic sugar

juangpc:
Hi, 
I wanted the functionality in the singleton EventManager to be accessible through static fcns. Right now it is set to the typical getInstance() method which then allows to retrieve public access to the class. Through this change (sugar) the api is directly accessible and it looks more natural to consume the class's API. 

+ bonus: In Communicator, there is a range for that was detaching from the Qt structure unnecessarily. Apparently, by making it const, c++ will iterate through references. Warnings-- (✊🏻)


juangpc:
@LorenzE did you told me you had any comments on this pr?

LorenzE (approved):
Sry that this approval took longer than expected. I tried a bit to rewrite this class via free functions (non member functions), but realised that in this specific case it is better to go with @juangpc implementation.

gabrielbmotta (approved):

PR #782, opened 2021-04-16
[ENH, WIP] MNE_Analyze Dynamically reload plugins

juangpc:
Hi everybody,
This allows to, while the mne_analyze app is running, click a menu "File>Reload Plugins" and if new dlls is found in the plugin folder, the menus, views, controls will be loaded. 

- This is not WASM ready. Not sure how you could modify the plugin set during wasm execution.

While doing this I noticed a few problems with the actual application itself, not directly related but I think my changes do improve the code. 
- PluginManager stores the plugins in a specific order given by a ```m_iOrder``` variable that works as a hint on what position you want your plugin to appear. Depending on which order you set this var to be, and what other plugins decide, your plugin will appear higher or lower in the menus. 
- Some warnings have been addressed. ✊🏻 
- If a plugin has already been loaded it will be not loaded again to the PluginManager list. 
- When data is loaded to the PluginManager there is new AbstractPlugin flag:```m_bInitialized``` flag that reflects that state.
- When the plugin has had its ```menus```, ```Controls``` and ```Views``` loaded to the menus of the application's window, there will be three member flags to show that: ```m_bMenuAlreadyLoaded```, ```m_bControlAlreadyLoaded``` and  ```m_bViewAlreadyLoaded ```.
- Added `.cpp` file to AbstractPlugin.
- MainWindow was a global object (within MNEANALYZE ns). It is now owned by AnalyzeCore, and there is a global pointer that gets intitialized and can be looking at MainWindow. This is used for the globalLogger.
- AnalyzeCore would previously give access to pluginmanager to the GUI. I see no reason why this should be the case, since these are two completely different things. So now MainWindow and PluginManager both communicate with their controller (AnalyzeCore). 
- I saw a few ```onThatEvent``` type of methods/slots... I think that we should try to adopt a more imperative style. The user (a gui, in this case, for instance) uses an api of another class... so you just use it. So not react to an event, just use the class you are supposed to use. (example: ```OnMainWindowClose``` -> ```close``` as in The gui uses the function ```close``` provided by the MainWindow Object).

If you check the files, you'll find a few changes in Plugins themselves. I solve a few tiny typo/style problems in a few of them.

Still a few things to do, like: 
-  PluginManagers could generate a leak of the created menu objects if ```getMenus()``` is called more than once. Although This is not the responsibility of AbstractPlugin, at least don't think so.


applications/mne_analyze/mne_analyze/mainwindow.cpp
```diff
@@ -77,57 +78,56 @@ using namespace DISPLIB;
 // DEFINE MEMBER METHODS
 //=============================================================================================================
 
-MainWindow::MainWindow(QSharedPointer<ANSHAREDLIB::PluginManager> pPluginManager,
-                       QWidget *parent)
+MainWindow::MainWindow(AnalyzeCore* pAnalyzeCore, QWidget *parent)
```
LorenzE, line 81: Our [coding conventions](https://mne-cpp.github.io/pages/development/contr_style.html) ask for aligning function parameters underneath each other.

applications/mne_analyze/mne_analyze/mainwindow.cpp
```diff
@@ -327,95 +438,21 @@ void MainWindow::createLogDockWindow()
...
-    for(AbstractPlugin* pPlugin : pPluginManager->getPlugins()) {
+    for(auto pPlugin : m_pAnalyzeCoreController->getLoadedPlugins()) {
         pPlugin->setObjectName(pPlugin->getName());
-        if(pPlugin) {
+        if(pPlugin && pPlugin->menuAlreadyLoaded() == false) {
             if (QMenu* pMenu = pPlugin->getMenu()) {
                 // Check if the menu already exists. If it does add the actions to the exisiting menu.
                 if(pMenu->title() == "File") {
                     for(QAction* pAction : pMenu->actions()) {
+#ifdef WASMBUILD
```
LorenzE, line 451: Please mind indentation of #ifdef just as you did above

applications/mne_analyze/mne_analyze/mainwindow.cpp
```diff
@@ -429,21 +466,28 @@ void MainWindow::createPluginMenus(QSharedPointer<ANSHAREDLIB::PluginManager> pP
...
+    {
+        setTabPosition(Qt::LeftDockWidgetArea,QTabWidget::West);
+        setTabPosition(Qt::RightDockWidgetArea,QTabWidget::East);
+        setDockOptions(QMainWindow::ForceTabbedDocks);
+    }
 
     //Add Plugin controls to the MainWindow
-    for(AbstractPlugin* pPlugin : pPluginManager->getPlugins()) {
-        if(QDockWidget* pControl = pPlugin->getControl()) {
+    for(AbstractPlugin* pPlugin :m_pAnalyzeCoreController->getLoadedPlugins()) {
```
LorenzE, line 487: add one space after `:`

applications/mne_analyze/mne_analyze/mainwindow.h
```diff
@@ -66,9 +68,9 @@ class QDockWidget;
 class QGridLayout;
 QT_END_NAMESPACE
 
-namespace ANSHAREDLIB {
-    class PluginManager;
-}
+//namespace ANSHAREDLIB {
+//    class PluginManager;
+//}
```
LorenzE, line 73: can this be removed?

applications/mne_analyze/mne_analyze/mainwindow.h
```diff
@@ -103,12 +105,17 @@ class MainWindow : public QMainWindow
     /**
      * Constructs a MainWindow which is a child of parent.
      *
+<<<<<<< HEAD
      * @param[in] pPluginManager Pointer to the plugin manager. It is needed to display subwindows froms plugins.
      * @param[in] parent Pointer to parent widget; If parent is Q_NULLPTR, the new MainWindow becomes a window.
+=======
+     * @param [in] pAnalyzeCore Pointer to the application controller class. It is needed to display subwindows froms plugins.
+     * @param [in] parent Pointer to parent widget; If parent is Q_NULLPTR, the new MainWindow becomes a window.
+>>>>>>> d7cfc0499 (rename log structure mainwindow stores analyze core not pluginmanager)
```
LorenzE, line 114: Remove git conflict indicators

LorenzE (changes requested):
Just some minor comments

gabrielbmotta (commented):
I'm running into an issue where the reloaded plugin's controls get added underneath the others instead of joining the stack of controls
pluginload

juangpc:
Thanks for checking this out. Interesting. I'll research about it.

applications/mne_analyze/libs/anShared/Management/pluginmanager.cpp
```diff
@@ -70,7 +72,7 @@ PluginManager::PluginManager(QObject *parent)
 
 PluginManager::~PluginManager()
 {
-    for(AbstractPlugin* plugin : m_qVecPlugins)
+    for(AbstractPlugin*& plugin : m_qVecPlugins)
```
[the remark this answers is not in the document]
juangpc: Agree, but I'd like to leave it as is. The addition of the reference just makes the compiler deal with references and avoids a warning.

applications/mne_analyze/libs/anShared/Plugins/abstractplugin.h
```diff
@@ -152,27 +159,102 @@ class ANSHAREDSHARED_EXPORT AbstractPlugin : public QObject
...
     //=========================================================================================================
     /**
-     * Initializes the plugin based on cmd line inputs given by the user.
+     * Called by the EventManager in case a subscribed-for Event has happened.
      *
+<<<<<<< HEAD
      * @param[in] sArguments  the cmd line arguments.
+=======
+     * @param e The Event that has taken place
+>>>>>>> 0ecfd5a34 (define qabstractplugin methods as non virtual when there is no need for that)
```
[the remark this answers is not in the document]
juangpc: done

juangpc (commented):
I think we've addressed all the comments.

gabrielbmotta (approved):

applications/mne_analyze/libs/anShared/Management/pluginmanager.cpp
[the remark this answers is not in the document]
LorenzE: is there also a warning for `atuo&`

LorenzE (approved):
sry I totally forgot about this.

PR #783, opened 2021-04-19
upload_url for coverity updated and curl command adapted.

jobehrens:


LorenzE:
@jobehrens Thanks for taking a look at this. Did you have a chance to test this on your end?

jobehrens:
@LorenzE Yes, I ran coverity locally and successfully committed the results that way (see: https://scan.coverity.com/projects/mne-tools-mne-cpp). Assuming that the coverity token is correctly stored in github, it should work like this.

LorenzE (approved):

LorenzE:
@jobehrens [This coverity workflow](https://github.com/mne-tools/mne-cpp/runs/2395906803?check_suite_focus=true) ran 11 hours ago and is not showing up on the Coverity website. Instead the latest analysis run is the one you did on 19.04.2021. Any idea what went wrong?

jobehrens:
I have changed one more small thing. Now it works. I had used only one email address for the test purpose, so I didn't notice the error before. I will make a new pull request.

PR #789, opened 2021-05-05
[ENH] MNE_Scan splashscreen

juangpc:
I make MNE Scan's splash screen hide 1 second after the main application's gui has finished its setup. During that second the splash screen can be hidden just by clicking on it, and since the splashscreen's management is offloaded to a new thread, even with the splash screen on top, the main GUI is still responsive. 

I think it gives a much sense of well tuned application. Since it is pretty much the first interaction with the user, i think it has some importance. 

More info... Some prs back, MNE Scan splash screen used to not be taken off by the application after it started. It would simply remain underneath the main gui window. I fixed that by making mne_scan's splash screen to hide the moment the initialization of the main screen had finished. Right before the main application loop started. This was not good enouth, and since then, I planned on finding time to implement the behavior shown in this pr.

thanks.

### DIFF
--- a/applications/mne_scan/mne_scan/main.cpp
+++ b/applications/mne_scan/mne_scan/main.cpp
@@ -171,8 +171,6 @@ int main(int argc, char *argv[])
     fmt.setSamples(10);
     QSurfaceFormat::setDefaultFormat(fmt);
 
-    mainWin.hideSplashScreen();
-
     int returnValue(app.exec());
 
     return returnValue;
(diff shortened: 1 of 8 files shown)


juangpc:
This PR is expected to be merged after the change to "main", so I think it is ok if it doesn't pass all the ci jobs at this point.

gabrielbmotta:
> This PR is expected to be merged after the change to "main", so I think it is ok if it doesn't pass all the ci jobs at this point.

From the look of the failing tests, there are some code errors that caused the tests to fail,

mainwindow.cpp:285:81: error: ‘MNESCAN::MainSplashScreen::SPtr {aka class QSharedPointer<MNESCAN::MainSplashScreen>}’ has no member named ‘get’

gabrielbmotta (approved):

gabrielbmotta (approved):

PR #790, opened 2021-05-05
[ENH] Avoid MNE_SCAN PluginItems' names in PluginScene to get chopped.

juangpc:
Hi, 

The size of each PluginItem in the PluginScene (PluginGui) in MNE_SCAN is a bit wider now. This makes it better to print a significant amount of characters of the name in the little plugin box. Before this, it was 8 characters per line in two lines. Now 
it is only one line with the first 10 characters. I think it is easier to remember and cleaner. 

It solves a bug, generated by some text being printed outside the box, which, when the box was dragged around would generate a little path on the PluginScene.

The complete name of the plugin remains visible in the control view once each PluginItem box is clicked. 

This is how it looks now: 
image

This is how it looks with this pr
image

### DIFF
--- a/applications/mne_scan/mne_scan/pluginitem.cpp
+++ b/applications/mne_scan/mne_scan/pluginitem.cpp
@@ -58,7 +58,7 @@ using namespace MNESCAN;
 PluginItem::PluginItem(SCSHAREDLIB::AbstractPlugin::SPtr pPlugin, QMenu *contextMenu, QGraphicsItem *parent)
 : QGraphicsPolygonItem(parent)
 , m_pPlugin(pPlugin)
-, m_iWidth(90)
+, m_iWidth(12 * pPlugin->getName().length())
 , m_iHeight(40)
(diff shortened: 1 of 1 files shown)


juangpc:
done. 

image

LorenzE:
Why are some boxes now longer than their respective name? For example, source localization's box is way too long.

juangpc:
> Why are some boxes now longer than their respective name? For example, source localization's box is way too long.

Because this is just a bad fix to a bad implementation. The purpose of this fix is to somewhat improve the plugin representation, it is not solved at all, but it is better than it is right now in main.

LorenzE:
Imho, these large boxes look worse than your solution before. Maybe go back to the initial fix of this PR where the size was hard coded?

juangpc:
> Imho, these large boxes look worse than your solution before. Maybe go back to the initial fix of this PR where the size was hard coded?

Yes, thank you. That's a good idea.

PR #794, opened 2021-05-18
[ENH] Tracer functionality in utils lib

juangpc:
This is a tracer class comes in handy whenever measuring executing times is needed.

It also might become very useful to understand execution flow of a complex algorithm or implementation. 

It traces multiple threads. 

It still needs documentation.


applications/mne_scan/mne_scan/main.cpp
```diff
@@ -71,8 +72,7 @@
 
 using namespace SCMEASLIB;
 using namespace MNESCAN;
-using namespace Eigen;
-using namespace UTILSLIB;
+//using namespace Eigen;
```
LorenzE, line 75: remove?

applications/mne_scan/plugins/hpi/hpi.cpp
```diff
@@ -1,589 +1,615 @@
-//=============================================================================================================
```
LorenzE, line 1: What happened here? Or is this just some git related change which is not actually change? Did you replace the files? Seems a lot of files in this PR are affected. Any way this can be reverted and only include the MNE_TRACE cahnges? It will overwrite the history and make using git blame difficult in the future.

libraries/utils/utils.pro
```diff
@@ -102,6 +104,13 @@ header_files.path = $${MNE_INSTALL_INCLUDE_DIR}/utils
 
 INSTALLS += header_files
 
+#contains(MNECPP_CONFIG, trace) {
+#    message("Building Utils lib with MNE_Tracer support.")
+#    DEFINES += TRACE
+#    SOURCES += mnetracer.cpp
+#    HEADERS += mnetracer.h
+#}
```
LorenzE, line 112: Why is this commented out?

mne-cpp.pri
```diff
@@ -79,6 +81,18 @@ contains(MNECPP_CONFIG, static) {
...
+    message("Building with MNE_Tracer support")
+    DEFINES += TRACE
+}
+
+#win32{
+#    QMAKE_CXXFLAGS += /FI "$$shell_path($${PWD}/libraries/utils/mnetracer.h)"
+#}
+#linux{
+#    QMAKE_CXXFLAGS += -include $$shell_path($${PWD}/libraries/utils/mnetracer.h)
+#}
```
LorenzE, line 94: Can you explain what this commented code does?

tools/python/tracer_macros.py
```diff
@@ -0,0 +1,99 @@
+import re
```
LorenzE, line 1: What does script exactly do?

libraries/inverse/inverse_global.h
```diff
@@ -42,6 +42,7 @@
 //=============================================================================================================
 
 #include <QtCore/qglobal.h>
+#include <utils/mnetracer.h>
```
LorenzE, line 45: why is this include in the inverse_global.h file? This way it will be included in a lot of headers files even if we do not use the mne tracer functionality. We should only include what we actually use.

LorenzE (changes requested):
Very cool @juangpc I have some questions. Maybe add a docu page as well?

juangpc:
After our conversation yesterday, I'd like to comment... 
In relation on how to structure this class in the project there are different options. 

(Option A)
One option is to include it as normal in utils lib. Drawback is that it has static members so the class will be executed even if unused. Only once... but still, it will affect start-up times, etc... I don't like it very much. But we're talking a few microsecs.
Good part is that it is simple and starightforward and the most usual way all classes are added to the project. And you can turn the tracer on or off with a define... pretty easy.

(Option B)
Another option is to include it in the utils library only if the trace flag is set. I like this option. When not needed, it saves the class from being included in the project compilation at all, not even executable. And it works fine in qtcreator. The main problem is that if there is some code with a macro the code will not compile. So if you use it you can't turn it on/off... And I like the macro to defined with the same flag in MNECPP_CONFIG. 
```
contains(MNECPP_CONFIG, trace) {
    message("Building Utils lib with MNE_Tracer support.")
    SOURCES += mnetracer.cpp
    HEADERS += mnetracer.h
}
```

(Option C)
We could create a DevelopmentUtils library? This libarary would only get linked when development.

(Option D)
Create a Release option which will not include MNETracer in the utils library. I'm not sure if this is supported by qmake:
```
contains(MNECPP_CONFIG, trace) {
    message("Building Utils lib with MNE_Tracer support.")
    SOURCES -= mnetracer.cpp
    HEADERS -= mnetracer.h
}
```

In relation to the include. It would be nice if the header file for MNETracer could be automatically included everywhere. 
I've experimented with this: 
```
win32{
    QMAKE_CXXFLAGS += /FI "$$shell_path($${PWD}/libraries/utils/mnetracer.h)"
}
linux{
    QMAKE_CXXFLAGS += -include $$shell_path($${PWD}/libraries/utils/mnetracer.h)
}
```
But the problem is that the force include options change within compilers and it doesn't work very well with the code-syntax highligher...  and I'd like to support gcc in linux, etc.... So I'm not satisfied at all.

Another option which Ive been considering is to have a global header file to be included in every project/plugin/application/whatever within mne-cpp. 
candidate --> mne.h <-- 
something like that... 

In there we could add some definitions, and whatever we need 
We could add macro flags like 
```c++
#ifndef release
#include <utils/mnetracer.h>
#endif
```
which would solve many of our previous problems. 

In need of neurons to make the right choice! help much appreciated @LorenzE  @gabrielbmotta  @RDoerfel 

Take care,

juangpc:
Apart from previous message and to set the configuration and example files to their final state... 
the pr is ready. 😃 
or so i think

applications/mne_scan/plugins/hpi/hpi.cpp
LorenzE: @juangpc Anyway you can get rid of these complete file replace changes? As I see it, the only changes to these files the MNE tracing macro calls.

LorenzE:
@juangpc Thanks for the detailed explanation! I would go for option A. It might not be the most elegant one performance wise, but I think it is the least invasive. Regarding including the tracer, I would try to only include it whenever you need it. There are even [tools ](https://include-what-you-use.org/) to explicitly look for superfluous includes and help you to remove them.

juangpc:
development changes reverted 
rebased
ready for merge

LorenzE:
@juangpc looks good! But why were the HPI files replaced? They show as changes but did not actually change, right? An interactive rebase might be useful to fix this?

juangpc:
It might be because I "reset" them to main . Il have a look.

juangpc:
> @juangpc looks good! But why were the HPI files replaced? They show as changed but did not actually change, right? An interactive rebase might be useful to fix this?

Don't see those changes. Think everything is ready for merge...

LorenzE (approved):

PR #798, opened 2021-05-28
[MAINT] BabyMEG plugin tweaks

gabrielbmotta:
- Remove blocking check for initialization of FiffInfo before it can be initialized.
- Remove 'Connect' tab and associated functions.

### DIFF
--- a/applications/mne_scan/plugins/babymeg/babymeg.cpp
+++ b/applications/mne_scan/plugins/babymeg/babymeg.cpp
@@ -199,13 +199,6 @@ bool BabyMEG::start()
 
     m_pMyClient->ConnectToBabyMEG();
 
-    if(!m_pFiffInfo) {
-        QMessageBox msgBox;
-        msgBox.setText("FiffInfo is missing!");
-        msgBox.exec();
-        return false;
-    }
-
     if(!m_pMyClient->isConnected()) {
         qInfo() << "BabyMEG::start - Not connected to BabyMEG device. Try connecting manually via the Connection tab.";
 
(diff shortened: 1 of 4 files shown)


juangpc (approved):

applications/mne_scan/plugins/babymeg/babymeg.cpp
```diff
@@ -199,13 +199,6 @@ bool BabyMEG::start()
 
     m_pMyClient->ConnectToBabyMEG();
 
-    if(!m_pFiffInfo) {
```
LorenzE, line 202: What happens if the fiff info has not been received yet?

LorenzE (commented):
I remember there were use cases where manually connecting to the babymeg via the GUI was helpful. Not sure which though ^^ Just wanted to mention that before you proceed with removing this.

applications/mne_scan/plugins/babymeg/babymeg.cpp
gabrielbmotta: When we were trying to connect yesterday, it has never been received at this point, it was only appeared set once the data started streaming.

juangpc: The way the code is right now, it is quite difficult to say with certainty. But, as far as I can see, by following the code, if m_pFiffInfo is not set, then the initConnector method would've done nothing which means that the rtmultisamplearray wouldn't have been set properly which means that the run method, basically, does nothing. So the plugin is stable while it leaves time for other plugins to feed info.
But what is clear is that if it ask for fiff info to be there at the beginning it almost-never works (because of timing).

PR #799, opened 2021-05-28
[MAINT] Add 'fits per second' control to hpi.

gabrielbmotta:
Preamble:
This is not a permanent fix by any means, this is a quick fix to have the continuous hpi in its current state more easily debugable during runtime (with different setups). This is a control for making iNumberOfFitsPerSecond variable during runtime, which was found to be a temporary solution to getting the continuous hpi fitting working on different incoming frequencies and data chunk sizes. iNumberOfFitsPerSecond does not set the number of fits per second, it is used to set the sample window used to perform the continuous fit, but the fit still seems to occur at whatever rate the data comes in. Treat this as a tuning variable, find a setting that works and leave it there. With that being said:

- Add controls for setting fits per seocnd in HpiSettingsView
- Make iNumberOfFitsPerSecond changeable during runtime

### DIFF
--- a/libraries/disp/viewers/hpisettingsview.cpp
+++ b/libraries/disp/viewers/hpisettingsview.cpp
@@ -95,6 +95,8 @@ HpiSettingsView::HpiSettingsView(const QString& sSettingsPath,
             this, &HpiSettingsView::compStatusChanged);
     connect(m_pUi->m_checkBox_continousHPI, &QCheckBox::clicked,
             this, &HpiSettingsView::contHpiStatusChanged);
+    connect(m_pUi->m_spinBox_fps, qOverload<int>(&QSpinBox::valueChanged),
+            this, &HpiSettingsView::fitsPerSecondChanged);
     connect(m_pUi->m_doubleSpinBox_maxHPIContinousDist, static_cast<void (QDoubleSpinBox::*)(double)>(&QDoubleSpinBox::valueChanged),
             this, &HpiSettingsView::allowedMeanErrorDistChanged);
     connect(m_pUi->m_doubleSpinBox_moveThreshold, static_cast<void (QDoubleSpinBox::*)(double)>(&QDoubleSpinBox::valueChanged),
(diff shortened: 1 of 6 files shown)


juangpc (dismissed):

juangpc:
It is great that it is saved in settings. 
One doubt. Since the variable ```m_iNumberOfFitsPerSecond``` doesn't really affect the actual number of fits per second done, shouldn't we change at least the control name in the gui?

gabrielbmotta:
> It is great that it is saved in settings.
> One doubt. Since the variable `m_iNumberOfFitsPerSecond` doesn't really affect the actual number of fits per second done, shouldn't we change at least the control name in the gui?

Sure, what do you propose?

juangpc:
Not sure, something like: fittingWindowSize or numSamplesToFit. to be set explicitly in samples by the (smooth) operator...

juangpc:
> Not sure, something like: fittingWindowSize or numSamplesToFit. to be set explicitly in samples by the (smooth) operator...

Let me push this. I'll do it.

juangpc (approved):

libraries/disp/viewers/hpisettingsview.cpp
```diff
@@ -242,6 +253,7 @@ void HpiSettingsView::loadSettings()
     m_pUi->m_checkBox_useComp->setChecked(settings.value(m_sSettingsPath + QString("/HpiSettingsView/useCOMP"), false).toBool());
     m_pUi->m_checkBox_continousHPI->setChecked(settings.value(m_sSettingsPath + QString("/HpiSettingsView/continousHPI"), false).toBool());
     m_pUi->m_doubleSpinBox_maxHPIContinousDist->setValue(settings.value(m_sSettingsPath + QString("/HpiSettingsView/maxError"), 10.0).toDouble());
+    m_pUi->m_spinBox_samplesToFit->setValue(settings.value(m_sSettingsPath + QString("/HpiSettingsView/fittingWindowSize"), 1000).toInt());
```
gabrielbmotta, line 256: It looks like here the default is set to 1000, whereas in the constexpr at the top of the file you set the default to 100. Is this intentional?

juangpc: Nope. I'll push a fix. It is a by product of me + the fact that some settings for a model are updated/initialized in the view!?

juangpc (dismissed):

juangpc (approved):

PR #806, opened 2021-06-07
[ENH] Create default FiffInfo for FtBuffer with minimal metadata

gabrielbmotta:
- Change logic to no longer expect (or need) neuromag extended header. Instead, default to minimal FiffInfo with made with data from simple ftbuffer header.

### DIFF
--- a/applications/mne_scan/plugins/ftbuffer/ftbuffer.cpp
+++ b/applications/mne_scan/plugins/ftbuffer/ftbuffer.cpp
@@ -259,7 +259,11 @@ bool FtBuffer::setupRTMSA(FIFFLIB::FiffInfo FiffInfo)
         return false;
     }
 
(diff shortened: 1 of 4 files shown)


gabrielbmotta:
Still need to test if this does not in fact break the existing parsing of neurmag extended header.

juangpc:
This is great @gabrielbmotta. It reminds me a lot of the way fieldtrip's python code reads the header ([👀](https://github.com/fieldtrip/fieldtrip/blob/463ffad122c0e27bd6b7d2bb5889319d4845c4ce/realtime/src/buffer/python/FieldTrip.py#L313)). 

Besides I can see the care for new atomic methods, renaming. Thanks. I think this is great code.

juangpc (approved):

PR #811, opened 2021-06-16
[FIX] MNE Analyze 

juangpc:
QMAKE_TARGET_DESCRIPTION  is what Qt uses to populate whatever field window's task manager uses to name the applications listed there. It makes sense to populate it with the name of the applications. 

Regarding the console. It is not needed when working with analyze in release mode.

### DIFF
--- a/applications/mne_analyze/mne_analyze/mne_analyze.pro
+++ b/applications/mne_analyze/mne_analyze/mne_analyze.pro
@@ -44,12 +44,11 @@ TEMPLATE = app
 QT += gui widgets network opengl svg concurrent charts
 qtHaveModule(printsupport): QT += printsupport
 
-CONFIG += console
-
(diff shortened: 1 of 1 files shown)


applications/mne_analyze/mne_analyze/mne_analyze.pro
```diff
@@ -37,17 +37,18 @@
 
 include(../../../mne-cpp.pri)
 
+QMAKE_TARGET_DESCRIPTION = A Framework for Electrophysiology.
```
gabrielbmotta, line 40: Should this say 'MNE Analyze' instead?

juangpc: 🥇 for me.

gabrielbmotta (approved):

PR #814, opened 2021-07-22
[FIX] LIB. Problem with static global variable inside the fiff library.

juangpc:
As mentioned in slack, there might be a problem with the life span of a static global variable in a library. 
This attempts to solve this issue, with a minimum as possible alteration to how the variable is working right now.


libraries/fiff/fiff_stream.h
```diff
@@ -653,7 +651,7 @@ class FIFFSHARED_EXPORT FiffStream : public QDataStream
      *
      * @return the position where the file id struct was written to.
      */
-    fiff_long_t write_id(fiff_int_t kind, const FiffId& id = defaultFiffId);
+    fiff_long_t write_id(fiff_int_t kind, const FiffId& id = FiffId::getDefault());
```
LorenzE, line 654: I do not really understand why we need a defaultFiffId variable in the first place? Why dont we just do the following:

`fiff_long_t write_id(fiff_int_t kind, const FiffId& id = FiffId());`

juangpc: The reason why I didn't propose that is because I want to keep, as much as humanly possible, the behavior of the software as it is right now. So, right now, there is one instance of an object that is the default object. Every call of ```write_id``` (with one parameter) will use exactly the same object as input. 
With the change you propose it would be a different instance of ```FiffId``` each call and I don't know (nor want to know) if there is some kind of assumption from the library or its users, that it is supposed to be just one an only ```defaultFiffId```. 
And, as you know, this all is my journey to leraning c++, so I think it all serves as good examples. So the question I ask myself (and you 😄) is:
_How can we solve the problem generated by this static object declared globally in a dynamically linked library, without modifying the rest of the code?_

mne-cpp-global.pri
```diff
@@ -0,0 +1,140 @@
+############################################## GLOBAL FUNCTIONS ###############################################
```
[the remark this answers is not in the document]
juangpc: I don't really remember. Somehow I see no reason to rename it now.

libraries/fiff/fiff_stream.h
LorenzE: The default ctor should always return an instance of the FiffId with the same state but different memory allocation. The only problem I see with the current implementation (the one in main) is that it is not efficient memory wise, because every translational unit which includes fiff_stream.h will have an instance of FiffId with the same state but different memory allocation. What you proposed fixes the memory issue, but basically mimics the behaviour of the default ctor. If you do not want to change too much of the current code base and still be memory efficient, I would specify defaultID as extern in the fiff_id.h and initialse it in fiff_id.cpp.

mne-cpp-global.pri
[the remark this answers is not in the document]
juangpc: it was a new file, not renamed. I think it was just some test or something.

libraries/fiff/fiff_stream.h
juangpc: > The default ctor should always return an instance of the FiffId with the same state but different memory allocation.

Couldn't agree more. 

> The only problem I see with the current implementation (the one in main) is that it is not efficient memory wise, because every translational unit which includes fiff_stream.h will have an instance of FiffId with the same state but different memory allocation.

I'm not sure if that is the only problem. When you load a linked lib, at run-time, you'll get an address of that global object ```defaultFiffId```. If you unload the dll, that address will be dangling. That is the main problem, that I see. And that is the warning underlined by clang (as I've mentioned before). (There is a link I shared with you a few days ago). 
Now, this that you mention, are you sure?  When loading a linked library, there are read-only memory pages that get shared between all the applications/libs that load the library. So if the compiler thinks that we're not modifying the instance then it will be in that read-only section, and shared among applications, so no memory inefficiency. However if it is not modified then we should just use the default constructor indeed, and move on. What I'm trying to say is that memory efficiency is really not my objective. What I'm after is to be able to, in the future, hot-swap, a dll during run-time and be sure that everything exists and is stable. I'm walking towards a situation where you can edit a plugin, while mne_scan is running, unload, and reload it, and check its behavior. So I want everything declared within the dll to be stable. 

> If you do not want to change too much of the current code base and still be memory efficient, I would specify defaultID as extern in the fiff_id.h and initialize it in fiff_id.cpp

What I want to take real care of is to not modify the behavior of the users of the classes that are already in the code. I'm just trying to figure out a solution considering the case where the api's could not be modified. In our case we probably modify everything we wanted (I think in this case it was used in 4 different occasions, so not big deal) but I want to solve it as if we couldn't afford a modification as perceived from the outside.

I don't understand how declaring defaultFiffId as extern would solve any of the problems, the instability nor the memory inefficiencies. Because the scope of the extern keyword would be app/lib bounded.

juangpc: Here is a good read about this. 

https://stackoverflow.com/questions/1538137/c-static-global-non-pod-theory-and-practice

LorenzE: Thanks for the detailed explanation @juangpc. I initially missed the point that you want to link explicitly. My explanation was assuming implicit linkage, which is what we currently do in mne-cpp.

Regarding your explicit linkage case. I can see how global variables in linked libraries could get problematic. Once the library is released, application and other lib code depending on released global variables are in a problematic state. I guess this just boils down to the fact that one should:

(a) try to avoid global variables in general
(b) make sure, if you use global variables, that they are only used in the same library or in libraries which are always loaded/unloaded together with the library. Our plugins are a good example for this. They are independent from each other and can be loaded/unloaded separately. First, I think that libraries should only expose an interface to the user, which the user needs to get the job done. Global variables should not be exposed to the user. Second, I think that a library should be able to do its job without depending on global variables of other libraries.

So if you want to make the, e.g., Fiff library to be loadable explicitly, you would need to refactor all global variables in that library, which are currently used outside of it. FiffId is not one of them since it is only used inside the Fiff library.

Regarding the memory efficiency of static global variables in libraries, you could be right. I just tried looking at it from a translational unit perspective. I do not know how efficient the compiler can handle/organize static global variables in the case of them being defined in a linked library. Interesting, I will investigate :)

juangpc: I learned that in a cppcon talk. 

My smiley means I agree, and I thank you for your comments. I agree with the (a) and (b) take home knowledge pills.

juangpc: https://docs.microsoft.com/en-us/previous-versions/dtba4t8b(v=vs.140)?redirectedfrom=MSDN

gabrielbmotta (approved):

PR #815, opened 2021-07-22
[FIX] ANONYMIZER Delete unnecessary const qualifier for signals

juangpc:
Qt signals should not carry a const qualifier.

### DIFF
--- a/applications/mne_anonymize/mainwindow.h
+++ b/applications/mne_anonymize/mainwindow.h
@@ -249,7 +249,7 @@ signals:
      * @param[in] s Absolute path of the input file.
      *
      */
-    void fileInChanged(const QString& s) const;
+    void fileInChanged(const QString& s);
 
     //=========================================================================================================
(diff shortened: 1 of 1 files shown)


gabrielbmotta (approved):
👍

applications/mne_anonymize/mainwindow.h
```diff
@@ -249,7 +249,7 @@ class MainWindow : public QMainWindow
      * @param[in] s Absolute path of the input file.
      *
      */
-    void fileInChanged(const QString& s) const;
```
LorenzE, line 252: @juangpc Can you elaborate why you feel the const qualifier is redundant here? I always find it nice to read which functions do not change the internal state of an instantiated object.

juangpc: Because they're signals.

LorenzE (approved):

PR #816, opened 2021-07-22
MAINT: Release script app bundles

gabrielbmotta:
- Build for Mac without app bundles

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -144,7 +144,7 @@ jobs:
     #     make install -j
     - name: Configure and compile MNE-CPP
       run: |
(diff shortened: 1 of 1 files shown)


juangpc (approved):

LorenzE:
@gabrielbmotta please check if macdeployqt still works on builds without app bundles.

juangpc (approved):

LorenzE:
@juangpc did you check if macdeployqt likes these changes you just merged? Otherwise we are left with a broken CI ...

LorenzE:
@juangpc @gabrielbmotta [Seems like macdeployqt does not like this](https://github.com/mne-tools/mne-cpp/runs/3144079371?check_suite_focus=true#step:9:12). There might be a way to configure macdeployqt to work on non app bundles though.

image

juangpc:
sorry about that.

LorenzE:
No worries. Any particular reason why we do not want to have app bundles in our releases. I am sure we can find a way to make the CI work without it. Depends on how much a priority this is for you.

PR #820, opened 2021-08-03
[WP][ENH] Ci linux using cl scripts

juangpc:
The objective of this PR is to, soon, be able to run, with exactly the same script, jobs locally and in the GitHub actions. There is still a bit of work to do.

 - [ ] Understand completely what do all the different GitHub workflows do.
 - [ ] Create a single multiplatform script for each workflow.
 - [ ] I'd like to discuss if we should just use python for all these and forgive us all of using bash scripting.
 - [ ] Standardize how to treat input arguments for all scripts. Source all the functions into one single script.
 - [ ] validate results.
 - [ ] ready to merge.


LorenzE:
@juangpc Is the goal of the PR to introduce a build script so that we can more easily build the project locally exactly just as we do in the CI? This would probably be a good idea since it will make the CI workflow yaml files easier to read.

I am just a bit confused because you mentioned "jobs" in the PR description. We should not try to replace every job in all the workflows by our own scripts. Workflows can house jobs, jobs are composed out of multiple steps, which again can be represented by Github actions. There are a lot of great Github actions available and we should not try to replace them with scripts written by us. Just take the Install Qt action as an example, rewriting this on our own is not a good idea. Not sure if this was your intention, but if it was we should discuss.

PR #822, opened 2021-08-06
[ENH] MNECPP - Comment installs keyword in pro files

juangpc:
The usage of ```INSTALLS``` keyword in some of the projects ```.pro``` files makes the project to show twice all the header files. 
That is the case in the library layer, for instance. 

Is there a reason why we're using ```INSTALLS``` keyword?

### DIFF
--- a/applications/mne_analyze/libs/anShared/anShared.pro
+++ b/applications/mne_analyze/libs/anShared/anShared.pro
@@ -124,12 +124,6 @@ INCLUDEPATH += $${EIGEN_INCLUDE_DIR}
 INCLUDEPATH += $${MNE_INCLUDE_DIR}
 INCLUDEPATH += $${MNE_ANALYZE_INCLUDE_DIR}
 
-# Install headers to include directory
-header_files.files = $${HEADERS}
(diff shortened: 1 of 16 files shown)


gabrielbmotta (dismissed):

gabrielbmotta:
I commented this in slack but I'll add it here as well:
>We seem to be using make install in the ci workflow release.yml, but in the deploy.bat script, we don't copy over the 'include' folder (which seems to be the default install location), so the copied header files don't get bundled with the release.

So as of right now it really doesn't seem like we are using these installed files.

LorenzE:
> I commented this in slack but I'll add it here as well:
> 
> > We seem to be using make install in the ci workflow release.yml, but in the deploy.bat script, we don't copy over the 'include' folder (which seems to be the default install location), so the copied header files don't get bundled with the release.
> 
> So as of right now it really doesn't seem like we are using these installed files.

Correct, currently we do not use them. This only gets interesting if we want to ship the libraries together with corresponding headers at some point. Feel free to remove the INSTALLS statement in all library .pro files.

juangpc:
This is ready ( I think).

(I hope).

LorenzE (approved):

PR #825, opened 2021-08-11
[FIX] Fix leaks in mne-c related libraries and few other classes

juangpc:
by coverity.

### DIFF
--- a/libraries/fwd/computeFwd/compute_fwd.cpp
+++ b/libraries/fwd/computeFwd/compute_fwd.cpp
@@ -2235,7 +2235,8 @@ void ComputeFwd::initFwd()
                                                  m_mri_head_t,
                                                  m_spaces,
                                                  m_iNSpace,out,m_pSettings->use_threads);
-        fclose(out);
+        if(out != Q_NULLPTR)
+            fclose(out);
     }
 }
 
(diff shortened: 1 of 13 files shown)


libraries/fwd/computeFwd/compute_fwd.cpp
```diff
@@ -2107,12 +2109,12 @@ void ComputeFwd::initFwd()
         if (!coilfile) {
             return;
         }
+        FREE_41(coilfile);
```
gabrielbmotta, line 2112: Why are we moving free coilfile to before we use it?

juangpc: Because on my quest for not leaking I forgot we're supposed to actually compute something, eventually...

juangpc: This code was updated and the actual commits lost in a rebase.

libraries/fwd/computeFwd/compute_fwd.cpp
```diff
@@ -1669,13 +1662,13 @@ bool mne_attach_env(const QString& name, const QString& command)
...
     FiffTag::SPtr t_pTag;
-    QList<FiffTag::SPtr> tags;
+//    QList<FiffTag::SPtr> tags;
     QFile fileInOut(name);
     FiffStream::SPtr t_pStreamInOut;
 
 //    if (fiff_new_file_id(&id) == FIFF_FAIL)
 //        return false;
-    id = FiffId::new_file_id();
+//    id = FiffId::new_file_id();
```
gabrielbmotta, line 1671: Variable 'id' gets used in line 1772 (former 1779)

gabrielbmotta (approved):

PR #826, opened 2021-08-16
[ENH,MAINT] Adding build timestamps

gabrielbmotta:
- Add build info header file to UTILS lib.
   - consexpr functions to return date/time and git hash

- Add build info to libraries/plugins
   - Add functions to *_global files for libraries and plugins.
   - namespace functions get build time info using UTILS build info header.
   - new *_global.cpp files to define functions.
   - Add rebuild dependency in .pro files so *_global.o get rebuilt whenever any of the library's files are changed.

- Add build info to apps
   - Add virtual function to abstract plugin in Scan and Analyze to query build info from loaded plugins
   - Add print step in plugin loading that shows plugin build info.
   - Add build info to main window 'About' page for main app build info.

Please excuse the size of the pr, it's 90% BSD headers / function documentation.


libraries/utils/utils.pro
```diff
@@ -131,3 +133,18 @@ contains(MNECPP_CONFIG, useFFTW):!contains(MNECPP_CONFIG, static) {
...
+
+win32 {
+force_update.commands += copy /b utils_global.h +,,
+
+}
+unix {
+force_update.commands += touch utils_global.h
+}
+
+PRE_TARGETDEPS += utils_global.h
```
juangpc, line 148: This is not compatible with msvsc.

juangpc:
It took me a while to figure out that ```PRE_TARGETDEPS``` is not compabible with msvc 🤦🏻 

I've modified the phony targets a bit.  All in 19b166d570394004405fe16845000d2216a07a8b

Could we go through this? 
Ideas:
Almost 200 files seems a bit too much. There has to be a better way.
With this implementation, if you want to create a plugin this means there is yet an additional thing you need to take care of?

Sorry but I don't understand why each plugin has to be modified in order to be extended. perhaps building an additional abstract class in between the pure abstract and each plugin. There are some pieces I don't remember right now. Would appreciate if we could go through this. 

Why does each plugin implement BUILD_TIME and BUILD_DATE functions? And why are they exported? But then in the actual implementation of the getBuildDateTime method mandated by the interface (```abstractPlugin```) it is directly the ```BUILDINFO::timestamp``` method used.

Averaging plugins in mne_scan and mne_analyze share the same namespace name. 

Thanks

juangpc:
Overall this has been a great exercise. Very useful for the future. 

This is a peek into how it looks. 
image

In the future, all applications could benefit of the experience gained with how mne_anonymize communicates with the buildsystem generator. 

However I still have a doubt: if you want a library or plugin to export a build_dateTime you need to create a phony target for that module/app/library/plugin or whatever it is. So that any modification to any of its files, triggers an update on the headerr file which updates the char* during compilation and thus updates the correct values. We have this implemented already for utils. 
Questions:
Is it implemented for the rest of the plugins?
Is it implemented for the rest of the applications?
... libraries?
@gabrielbmotta ? 

Apart from this, I think this is ready for merging.

juangpc:
- rename methods to more c++ish
 - replicate qmake update thingy for all plugins/libraries as it is in utils.
 - live a happy life

juangpc:
@gabrielbmotta  do we still need the getBuildDateTime() virtual method and its overrides in each inheritor of abstractPlugin? Can't we solve the issue now just with the ```PLUGIN_NAMESPACE::buildDateTime()``` ?

gabrielbmotta:
> @gabrielbmotta  do we still need the getBuildDateTime() virtual method and its overrides in each inheritor of abstractPlugin? Can't we solve the issue now just with the ```PLUGIN_NAMESPACE::buildDateTime()``` ?

That function is there to be called when loading in the plugins and displayed after the already existing loading messages. Are you proposing we use the namespace functions instead?

applications/mne_analyze/plugins/averaging/averaging.pro
```diff
@@ -130,3 +131,20 @@ contains(MNECPP_CONFIG, useFFTW):!contains(MNECPP_CONFIG, static) {
                 -lfftw3_threads \
     }
 }
+
+################################################## BUILD TIMESTAMP/HASH UPDATER ############################################
```
LorenzE, line 135: Can we put the code below into a function inside the mne-cpp.pri file? Would save a lo of redundant code.

juangpc: not sure, there are some variables that are context dependent. but if you know and want to try, go ahead. just give me a heads up before pushing.

juangpc:
ready for review.

juangpc:
Finally, I think this is ready now. Working in windows, mac and linux. 

Overall it was a good exercise. Learned a lot. However rather unsatisfied by the lack of flexibility of qmake. After a lot of care to not "hack", I couldn't find any other way apart from hacking my way through qmake so that make did what I wanted.

juangpc (approved):

juangpc (approved):
Thanks Gabriel for this last push.

PR #833, opened 2021-09-07
[FIX] Leak mne cov matrix

juangpc:


juangpc:
This should be ready for review

libraries/mne/c/mne_cov_matrix.cpp
```diff
@@ -162,10 +177,10 @@ int mne_decompose_eigen (double *mat,
     double *dmat = MALLOC_30(np,double);
     float  *vecp = vectors[0];
 
-    const char   *uplo  = "U";
-    const char   *compz = "V";
+//    const char   *uplo  = "U";
+//    const char   *compz = "V";
```
gabrielbmotta, line 181: If these are unused, should we just delete them entirely instead of commenting them out?

juangpc: agree

gabrielbmotta (approved):
👍

PR #835, opened 2021-09-10
[ENH] HPI Fit 3D View - multi view display

gabrielbmotta:
- Add 3 new cameras and viewports to 3dview (toggleable).
- Add controls to control3dview to select between single and multiview
- Make Multiview stack horizontally/vertically based on a available space and aspect ratio.

Here's what it looks like (disregard fit, it's sample_audvis_raw) 

multiview1

multiview2


gabrielbmotta:
Right now I have the three views showing front, side, and top. Top/bottom seems the least useful of the views, should we keep it? Should we do just front + side. Or front + side + some other view?

LorenzE (approved):
very cool!

LorenzE (approved):

PR #836, opened 2021-09-11
[ENH,Fix] (Hopefully) Solve occasional crashing of MNE Scan 

RDoerfel:
I observed occasional crashes of MNE Scan after the first successful fit. Specifically, the crash happened while the computed dev_head_trans was applied to the scaled and transformed average head. It seems that the issue was that the applied transformation was not affine, meaning that the last row was not (0,0,0,1). Qt only applies affine transformations and therefore errored sometimes. 

image

Since we should only apply affine transformations anyway, I implemented a new convenience function to set a new transformation of an existing FiffCoordTrans. This function enforces affine transformations automatically. The function also avoids working on members directly, which shouldn't be done in the first place as well, I guess.


LorenzE:
@RDoerfel Have you decided on whether to keep the setTransform() function?

RDoerfel:
> @RDoerfel Have you decided on whether to keep the setTransform() function?

I was thinking about adding a flag to FiffCoordTrans::make() which defaults to false. If it is set to true, it is forced to be affine. With that, we should not change any existing behavior, but add the possibility for the future? Of course, I would prefer to have it the other way around. So that affine transformations are the default. But this might be too invasive?

LorenzE:
Why not adding a function called makeAffine? I would be ok with adding a flag `forceAffine` to the make() function as well and set it to false as default.

LorenzE:
@gabrielbmotta @juangpc any preferences from your side?

gabrielbmotta:
> @gabrielbmotta @juangpc any preferences from your side?

I'd go with creating a new function 'makeAffine.' I'd hazard a guess that @juangpc would not like the second option, as he does not like default values. 😁

RDoerfel:
> > @gabrielbmotta @juangpc any preferences from your side?
> 
> I'd go with creating a new function 'makeAffine.' I'd hazard a guess that @juangpc would not like the second option, as he does not like default values. 😁

Of course, we can do that. But then we would add a new kind of function again. Wouldn't that be that same situation as in the beginning, just named differently?

juangpc:
Hi, 
Yes @gabrielbmotta knows me well. 😄 .
I would like to make things as easy as possible. For that matter, and since we all use ide today, I'd prefer a specific ```makeAfine``` fcn. If you don't like it, it is my opinion that it is better to override than use default values. I know that they're not exactly the same thing, but regardless, iit is better to double dereference a pointer than to go back and forth between header and implementation files.

All opinions are just that, most certainly mine. At the moment I most strongly believe that we should have a writing style that makes c++ more expressive and easy to follow. And that if we are not comfortable with our style, we should change it.

Would you remind me why did you change cols for rows in fitHPI fcn?

RDoerfel:
> Hi,
> Yes @gabrielbmotta knows me well. 😄 .
> I would like to make things as easy as possible. For that matter, and since we all use ide today, I'd prefer a specific `makeAfine` fcn. If you don't like it, it is my opinion that it is better to override than use default values. I know that they're not exactly the same thing, but regardless, iit is better to double dereference a pointer than to go back and forth between header and implementation files.

Regarding `makeAffine` see my previous updated commend. I think I would be fine with that, but our transformation might not necessarily affine. It is more about enforcing this 3x3 matrix + 3x1 vector part, which I think is the way it was supposed to be anyway. 

> All opinions are just that, most certainly mine. At the moment I most strongly believe that we should have a writing style that makes c++ more expressive and easy to follow. And that if we are not comfortable with our style, we should change it.

What do you mean by 'writing style'?

> Would you remind me why did you change cols for rows in fitHPI fcn?

Before I used `FiffCoordTrans::apply_trans()` the multiplication was carried out there. For that purpose, the matrix had to be transformed.  `FiffCoordTrans::apply_trans()` already takes care for that

juangpc:
> our transformation might not necessarily affine

Ok I see what you mean, that is true. So lets not call it affine then, neither in the fcn name neither in bool parameter, unless we actively check for "affinity" properties and validate them. If it is this extra row that we're adding then we can call it "standardize" or whatever you feel like calling it. Still, for our discussion we can use makeTransformationStandard  or something like that. 

Writing style? I'm sure you know what I mean. The set of rules to help write more easy-to-understand-and-follow code. Like, for instance.... avoid using default values, rather than overriding functions and let c++'s engine encapsulate the decision on what function to call.

RDoerfel:
> > our transformation might not necessarily affine
> 
> Ok I see what you mean, that is true. So lets not call it affine then, neither in the fcn name neither in bool parameter, unless we actively check for "affinity" properties and validate them. If it is this extra row that we're adding then we can call it "standardize" or whatever you feel like calling it. Still, for our discussion we can use makeTransformationStandard or something like that.
> 
> Writing style? I'm sure you know what I mean. The set of rules to help write more easy-to-understand-and-follow code. Like, for instance.... avoid using default values, rather than overriding functions and let c++'s engine encapsulate the decision on what function to call.

All right. Would you agree with an additional function `isStanderd()` to check if the current transformation is a standard transform? The ew could check before passing to qt3d and avoid possible crashes? Or is this already overengineering.

juangpc:
The way I see it, we should try to encapsulate. Make the makeTranform
function check if it is standard or not, and solve it or act
accordingly, or have two functions with same name and different signature
that take care of this matters internally. The user of the transform class
is probably worried about something else so I'd try not to bug too much
about it.

I don't think it is overengineering but an opportunity to think about
code, experiment and learn.

Juan

RDoerfel:
> @juangpc @RDoerfel Ready to merge? :)

Not yet. I forgot to discuss this at our last meeting. I'm a bit puzzled about what would be the best solution here. I agree that something like `setTransform` would be different from how it was done before, and might therefore not be optimal. However, I don't think that adding `makeStandard` in addition to `make` would be a good solution either. But I also understand that adding a flag is not a nice solution either :D

PR #838, opened 2021-09-18
[FIX] $ingle dollar variable dereferencing in pro files

juangpc:
Hi, after the work in #826, I've noticed a few places where single ```$``` are used, where probably (maybe) double dollar ```$$``` is required. Not sure. Maybe not. 

So if it correct as it is. Let's drop this and move on. If it is wrong, lets evaluate fixes.

### DIFF
--- a/applications/mne_analyze/mne_analyze/mne_analyze.pro
+++ b/applications/mne_analyze/mne_analyze/mne_analyze.pro
@@ -159,7 +159,7 @@ win32 {
 }
 
 unix:!macx {
-    QMAKE_RPATHDIR += $ORIGIN/../lib
+    QMAKE_RPATHDIR += $$ORIGIN/../lib
 }
 
 macx {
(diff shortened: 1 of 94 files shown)


juangpc:
Other places: 
Check ```COPY_DIR```, ```first```. Is this working?
https://github.com/mne-tools/mne-cpp/blob/6dfbef0e641e413f9f1d51db2e00ae8f10af67b5/applications/mne_scan/plugins/hpi/hpi.pro#L136
The Makefile looks like this, which I'm not sure is correct: 
image

Also, in line https://github.com/mne-tools/mne-cpp/blob/6dfbef0e641e413f9f1d51db2e00ae8f10af67b5/applications/mne_scan/plugins/hpi/hpi.pro#L140

```first``` is already a target by default isn't it?

Finally in ```QMAKE_FILE_OUT``` : 
https://github.com/mne-tools/mne-cpp/blob/6dfbef0e641e413f9f1d51db2e00ae8f10af67b5/tools/ifw_installer/windows/windows.pro#L64

gblame says 6 years ago. I see this is gitignored. How do we use this installer @LorenzE, would you help me understand it?

Thanks.

LorenzE:
@juangpc I never really used the installer. IMHO It is deprecated and can be removed.

juangpc:
Thanks. We discussed it and we decided some cases seemed OK, and others were OK to leave as they were. it was better to move on. 

But thanks.

PR #841, opened 2021-09-23
[FIX] Use correct path to hpi preset

gabrielbmotta:
- Change path from binary path to the path of the directory the binary is in.


LorenzE (approved):

juangpc:
Good fix.

PR #843, opened 2021-09-27
[Fix] Memory leak in MNE library in bem related functions

juangpc:
In ```add_bem_surface``` there seems to be a leak of ```newSurf``` 

In ```add_replace_display_surface``` I'm not sure about ```surf``` variable. @LorenzE do you remember about this function?

### DIFF
--- a/libraries/mne/c/mne_msh_display_surface_set.cpp
+++ b/libraries/mne/c/mne_msh_display_surface_set.cpp
@@ -427,7 +427,7 @@ int MneMshDisplaySurfaceSet::add_bem_surface(MneMshDisplaySurfaceSet* surfs,
         if (std::fabs(sum - 1.0) > 1e-4) {
             fprintf(stderr, "%s surface is not closed "
                                  "(sum of solid angles = %g * 4*PI).",name,sum);
-            return FAIL;
+            goto bad;
         }
     }
 
@@ -449,7 +449,14 @@ int MneMshDisplaySurfaceSet::add_bem_surface(MneMshDisplaySurfaceSet* surfs,
     return OK;
 
 bad : {
-        delete surf;
+        if(surf)
+        {
+            delete surf;
+        }
+        if(newSurf)
+        {
+            delete newSurf;
+        }
         return FAIL;
     }
 }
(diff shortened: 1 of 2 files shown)


libraries/mne/c/mne_msh_display_surface_set.cpp
```diff
@@ -442,21 +442,14 @@ int MneMshDisplaySurfaceSet::add_bem_surface(MneMshDisplaySurfaceSet* surfs,
...
 
 bad : {
-        if(surf)
-        {
-            delete surf;
-        }
-        if(newSurf)
-        {
-            delete newSurf;
-        }
```
LorenzE: Why are we not cleaning up newSurf anymore?

libraries/mne/c/mne_msh_display_surface_set.cpp
```diff
@@ -427,7 +427,7 @@ int MneMshDisplaySurfaceSet::add_bem_surface(MneMshDisplaySurfaceSet* surfs,
         if (std::fabs(sum - 1.0) > 1e-4) {
             fprintf(stderr, "%s surface is not closed "
                                  "(sum of solid angles = %g * 4*PI).",name,sum);
-            goto bad;
+            return FAIL;
```
LorenzE: if we go do not go to bad here, we do not clean up newSurf.

libraries/mne/c/mne_msh_display_surface_set.cpp
juangpc: Yes. We're on that. This  is WP. 
But before getting on with it, I'd really need some help trying to understand what is going on with ```add_repace_display_surface``` which also (probably) has a leak. I'd like to solve both issues at the same time. 
Could you help me understand what is this fcn doing?

here: https://github.com/mne-tools/mne-cpp/blob/edffebb123ce116e5bd1ef027acb8a9a217cb8c0/libraries/mne/c/mne_msh_display_surface_set.cpp#L459

LorenzE: I guess we parse through a set of display surfaces and if we find a match (based on the ID) we delete the old surface and replace it with a new one?

juangpc: Thank you.

juangpc:
@LorenzE , ready for review.

gabrielbmotta (dismissed):

gabrielbmotta (approved):

PR #850, opened 2021-10-06
[ENH] Replace fprintf(stderr, for printf(

juangpc:
Hi, I thin that the use of stderr in an application like any of the mne-scan is a bit outdated. We should provide a good logging mechanism eventually, but in the mean time, we can at least try to have all the messages printed to the same console and then start from there.


juangpc:
> bout having a good logging mechanism. But why is writing to different output streams outdated? If now output file is provided stderr will print to the same output and you can see the messages just as if you would with printf(), no?

it is outdated for many reasons. I'm even a bit surprised you want to go over this. The difference between stdout and stderr is just a convention *at the operating system level* with which a developer and a user agree that some information printed to the user is normal or "standard", and some other is related to diagnostic or debugging being performed by the suer. That was originally the case and unix and bourne shell. 
We have dlls so we could have different applications by default printing the same output to error output. Both standard and error outputs are unsynchronized so it is difficult to follow between them. Today, I'd like to have different applications writing to different logging mechanisms, and not having to look at two different locations to understand what is going on. 

But in any case, if you agree with the new logging tool. Lets keep this change which will in short time be replaced with a different tool altogether.

juangpc:
@LorenzE Please review this whenever you can.

LorenzE (approved):

PR #854, opened 2021-10-25
[FIX] Initialize value in icp function

juangpc:
There seems to be a use of an uninitialized value in icp.cpp


juangpc:
@gabrielbmotta Please review this when you can.

gabrielbmotta (approved):

PR #859, opened 2021-11-02
[MAINT] Stability improvements to stopping measurements in FtBuffer and MNE Scan

gabrielbmotta:
- Add outs for loops where fetching data could get stuck.
- Change order of function calls when ending measurements.

### DIFF
--- a/applications/mne_scan/plugins/ftbuffer/ftbuffer.cpp
+++ b/applications/mne_scan/plugins/ftbuffer/ftbuffer.cpp
@@ -148,8 +148,13 @@ bool FtBuffer::stop()
 
     //stops separate producer/client thread first
     m_pProducerThread.requestInterruption();
+    int itries = 0;
     while(m_pProducerThread.isRunning()) {
         msleep(10);
+        if(itries > 10){
+            break;
+        }
+        ++itries;
     }
 
     //Reset ftproducer and sample received list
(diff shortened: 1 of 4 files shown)


juangpc (dismissed):

LorenzE (dismissed):
Let's discuss this in our meeting tomorrow. This stuff enters the "danger zone" ;)

applications/mne_scan/plugins/ftbuffer/ftbuffer.cpp
```diff
@@ -146,6 +146,9 @@ bool FtBuffer::stop()
 
     m_bIsConfigured = false;
 
+    requestInterruption();
+    wait(500);
```
LorenzE, line 150: Can you try if this also works without the 500ms wait?

gabrielbmotta: It seems to work fine without the wait command. Escaping the main run loop is pretty fast, so maybe it doesn't affect it with what we are doing now. The [wait call](https://doc.qt.io/qt-5/qthread.html#wait) does stop blocking early if the function returns quickly (which it seems to).

juangpc:
What state this is in? ready for merging? 
I have a couple of questions. Perhaps during the day.

juangpc (approved):

PR #860, opened 2021-11-02
[MAINT] Clear measurement info when stopping hpi plugin

gabrielbmotta:
- Cleared saved Fiff Info and Digitizer Data when resetting plugin state.


juangpc (approved):

LorenzE (dismissed):

applications/mne_scan/plugins/hpi/hpi.cpp
```diff
@@ -153,6 +153,7 @@ bool Hpi::start()
 
 bool Hpi::stop()
 {
+    requestInterruption();
```
LorenzE, line 156: IF you ask for a request interruption here, you would need to include a check for interruption in the run() method.

applications/mne_scan/plugins/hpi/hpi.cpp
```diff
@@ -163,6 +164,9 @@ void Hpi::resetState()
 {
     m_bPluginControlWidgetsInit = false;
     m_pCircularBuffer->clear();
+
+    m_pFiffInfo.clear();
```
LorenzE, line 168: When is the resetState() called. Please be aware that all plugins share one single fiff info. So clearing it here will also clear it for all other plugins.

juangpc: Wow.

gabrielbmotta: I am clearing the pointer instance, which just drops the reference by one.

applications/mne_scan/plugins/hpi/hpi.cpp
gabrielbmotta: The loop in the run method of the hpi fit is `while(!isInterruptionRequested())`. Do you mean add more break points with that condition inside the loop?

applications/mne_scan/plugins/hpi/hpi.cpp
LorenzE: @gabrielbmotta ah ok. Now I see it too. That should be fine then :)
@juangpc Comments like "wow." are not very expressive. What would you like to say here?

applications/mne_scan/plugins/hpi/hpi.cpp
LorenzE: sry my bad. isInterruptionRequested() is used already indeed.

applications/mne_scan/plugins/hpi/hpi.cpp
juangpc: Yes. Sry for the noise. Deleted.

LorenzE (approved):

juangpc (approved):

PR #863, opened 2021-11-07
FIX: Remove goto and return false directly.

jobehrens:
(Coverity 159841 Logically dead code)

### DIFF
--- a/libraries/inverse/dipoleFit/ecd_set.cpp
+++ b/libraries/inverse/dipoleFit/ecd_set.cpp
@@ -247,7 +247,7 @@ bool ECDSet::save_dipoles_bdip(const QString& fileName)
     if (fclose(out) != 0) {
         out = NULL;
         printf(fileName.toUtf8().data());
-        goto bad;
+        return false;
     }
     fprintf(stderr,"Save %d dipoles in bdip format to %s\n",nsave,fileName.toUtf8().data());
     return true;


juangpc:
Hi @jobehrens, to leave the function through a ```goto bad``` is a way of approaching the equivalent of a c-ish destructor. Once you get used to mne-c's code style, this kind of organization is all over the place. So, I see your point but I'd prefer to keep the function executing the ```bad``` section. 

Since there is a codecov problem regarding this pr, perhaps you can set the problem as false positive. @LorenzE @gabrielbmotta, what do you think? 

As a side note, I would agree that ```bad``` is a bad name for such a "wrap-up" section, but that would be a different discussion.

LorenzE:
@jobehrens @juangpc I just realized that skipping the goto command would lead us to not call `unlink()`, which might be problematic?! Not sure if the file created must be removed in case of failure. In general I am ok with removing or skipping these goto commands, if cleanup is properly handled/refactored.

jobehrens:
Hi @juangpc and @LorenzE. In line 248 out = NULL is set. Therefore only false is returned anyway, because the condition in line 256 is never fulfilled. That's why I removed the goto bad there, to make it clear that nothing happens there anyway. I'm not sure if the unlink is useful if there was an error on the close.

LorenzE:
@jobehrens You are right. This can be merged as is from my side. @juangpc what do you think?

juangpc:
I still think that RAII is a good idea, including mne-c's version of it. So I still think this is sort of a false discover by the static tool. (Probably due to the fact that mne-c is unorthodox, in its style.)
But I also think think this is not that important at all. 
About to merge.

LorenzE:
RAII is of course a good idea, no one questioned that. But here the code simply does not perform it, even if we do call `goto bad` in line 250. It will still only return false as @jobehrens mentioned before. So I think the static code analysis tool does exactly what it is supposed to do: Catching situations where the code complexity can be reduced and readability improved. For someone reading over the code, I would imagine reading `return false` rather than `goto bad` (plus understanding the bad section) is much easier.

juangpc:
This is not so important in my opinion, so I'm not sure why we are discussing this. But, I respectfully disagree @LorenzE. And probably I should explain myself better, but consider my comments as positive and well intentioned. Not only I've already merged this, but I also think it is fine. And even I didn't I also trust your opinion. So let me try to explain myself better.

The ```bad``` section should be executed even if in its current state, it does nothing. That is (partially) RAII. So that if in the future, someone edits the function in way that requires clean-up, he/she can add whatever is needed to the bad section and be sure that the clean-up will be always called. And the more alternative ```return``` statements we add to the function, the more move the code away from this RAII-like structure. That is why I mentioned that I _still_ like RAII.
Let me add that possibly, I have removed some of these sections in past pr, well, maybe I shouldn't have. But again, don't think this is such a big deal.
Also, I respectfully disagree with your statement regarding the static analysis tool. Not because you have an empty destructor, you should see a warning. (**at least before c++11, with the inclusion of ```default``` which is a different subject). I guess that it is unorthodox to have such function implemented with a code section, and so it is in my humble opinion, a false positive.
And finally, I also disagree with your last statement. I think that if you allocate heap and you just delete it in the ```bad``` section, just in case something goes wrong, etc... it all reduces code complexity and makes it easier for the developer to organize and handle different things that can happen. In this sense, we could also consider the bad section as an unorthodox exception handling structure (in c code!). Which, even if it is unused at this point in the development, still should be kept in use and ready for future modifications.

I do agree very much with you in the fact that explicit code is better than implicit structures that require de developer to understand this meta-code structures. And, in case you're interested, that is exactly what I shared with the author of the code when I had the chance. So yeah, ```return false``` is much easier to read and follow in that sense. But after some time spent with mne-c code (and descendants 😄 ), once you get this style with sections, I think it should be kept as is.

Thanks for reading such a long comment.

PR #873, opened 2021-12-02
[FIX] Add correct check for clang.... again

juangpc:
😮‍💨

### DIFF
--- a/applications/mne_anonymize/mne_anonymize.pro
+++ b/applications/mne_anonymize/mne_anonymize.pro
@@ -123,7 +123,7 @@ HEADERS += \
 FORMS += \
     mainwindow.ui
 
-*-clang++ {
+clang {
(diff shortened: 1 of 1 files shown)


applications/mne_anonymize/mne_anonymize.pro
```diff
@@ -123,7 +123,7 @@ HEADERS += \
 FORMS += \
     mainwindow.ui
 
-*-clang++ {
+clang {
```
LorenzE, line 126: why is this needed in only this one file? Weren't other files changed as well?

LorenzE (approved):

applications/mne_anonymize/mne_anonymize.pro
juangpc: all other 109 appearances where changed in #872 . This one escaped.

PR #875, opened 2021-12-21
FIX: Coverity 221388 Incorrect deallocator used

jobehrens:
221388 Incorrect deallocator used


juangpc (commented):
@jobehrens  thanks for this. 

I've silenced all other warnings in the file too. (We have too many warnings). 

@gabrielbmotta could you review and approve/comment?

libraries/inverse/dipoleFit/dipole_fit_data.cpp
```diff
@@ -758,10 +758,10 @@ int mne_decompose_eigen_3(double *mat,
     double *dmat = MALLOC_3(np,double);
     float  *vecp = vectors[0];
 
-    const char   *uplo  = "U";
-    const char   *compz = "V";
+    __attribute__((unused)) const char   *uplo  = "U";
```
gabrielbmotta, line 761: 2 Things about this:
- If these are local values that are not used, why not simply remove them instead of declaring them unused?
- Does this compile on windows? I thought this was gcc specific syntax.

juangpc: true. I checked, gcc and clang, but msvc does not. 
Today I learned about https://en.cppreference.com/w/cpp/language/attributes/maybe_unused... 
Interesting... although ahead of us. 😄  
Regarding why not deleting.... because there are some in/out paramaters which are not used and to modify them I would need to modify also the actual function calls. I it didn't seem to me so important. 
I'll roll back. (and do it right).

gabrielbmotta: Yes, I understand not changing the in/out, I was asking about using it on the local variables. Anyway, just some suggestions: Qt has Q_UNUSED that fills the same role, or we could cast to void.

juangpc: I casted to void, I'm about to push changes.
I've also reviewed all errors with clang-tidy and it seems there was another leak and a s* load of new warnings... so I'm settling.

libraries/inverse/dipoleFit/dipole_fit_data.cpp
```diff
@@ -605,7 +605,9 @@ void mne_free_cov_3(MneCovMatrix* c)
     if (c->sss)
         delete c->sss;
     c->bads.clear();
-    FREE_3(c);
```
gabrielbmotta, line 608: Are we sure this other matrix that was changed from Free3() to delete uses new? The source code in mne_cov_matrix still uses a lot of malloc. Is this another coverity identified issue?

gabrielbmotta (dismissed):

gabrielbmotta (approved):

PR #881, opened 2022-01-11
[MAINT] Utils library Qt-less refactor (Part 1 of many)

gabrielbmotta:
This goes over a majority of the utils library and adds a std lib api (and a std lib backend) to most classes and static functions.

- Overloaded functions with Qt-type parameters and return values with std lib versions.
- Changes internal logic where applicable to use std lib containers/classes
- When faced with functions with no input parameters and a qt return type, overloaded added and commented out.

Also:

- ~~min c++ version set to c++ 17 for use of filesystem library. Is this ok? this is a 5 year old standard at this point.~~
(reverted, was causing issues with ubuntu 18 and osx 10.15 and lower std lib version. No longer using filesystem lib. )

- Because adding more code results in a lower code coverage percentage even though the functions I overloaded were not being tested anyway (I think), I also wrote a test for circular buffer to bump up code coverage.

What is missing:
- Going over threading use of QtConcurrent::mapReduced in spectrogram.h/cpp. Somewhat relates to the now-dormant time frequency PR #785 
- QSemaphore use in circular buffer. Std lib of semaphore is c++20, and that is perhaps too recent a cut off point. We could implement our own semaphore with the same api as the c++20 version to make later replacement easy.

### DIFF
--- a/examples/ex_hpiFit/main.cpp
+++ b/examples/ex_hpiFit/main.cpp
@@ -272,6 +272,6 @@ int main(int argc, char *argv[])
             << "Average Duration:" << matPosition.col(9).mean() << "ms";
 
     if(bSave) {
-        IOUtils::write_eigen_matrix(matPosition, QCoreApplication::applicationDirPath() + "/MNE-sample-data/" + sNameOut);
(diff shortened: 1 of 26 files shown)


juangpc:
great!

gabrielbmotta:
The qt installer action for linux is failing.

juangpc (approved):

PR #882, opened 2022-01-20
[ENH] MNE Scan: change plugin gui items

gabrielbmotta:
- Sets the bounding rectangle around the text based on the size of the printed text with the used QPainter object.

#### For reference:

Plugin GUI now             |  Plugin GUI in this PR
:-------------------------:|:-------------------------:
plugins_old  |  plugins

Let me know if more/less spacing would look better. Or if we should look at doing line breaks on whitespace.


juangpc (approved):

juangpc:
resizeAsRectangle is not going to be a very fast method, but I do like the end result. Thanks!

LorenzE (approved):

PR #884, opened 2022-01-27
WIP [Maint] Refactoring Hpi Fitting

RDoerfel:
This PR aims to refactor the HPI class to make the code more readable, separate concerns, and make it easier to test. So far, no new functionality itself was implemented. I guess, most of the new lines are unit tests.

Following is a list of newly introduced classes. 

### HpiDataUpdater
Extracting important information like channel list from the FiffInfo. The used channels directly influence the data and projectors used for the HPI fit. However, this preparation based on the channel list has nothing to do with the HPI fitting algorithm and was therefore moved outside the HpiFit Class. Further, it provides the current SensorSet.

### SensorSet and SensorSetCreator
`SensorSet` is a class containing the MEG sensor information and is constructed by `SensorSetCreator`. `SensorSetCreator` stores the link to the coil-definition file and a general FwdCoil object which contains information about all possible sensor types, and therefor, no information that should be carried by a specific `SensorSet`

### HpiModelParameter
Is an object containing all the information (hpi freqs, sfreq, linefreq, number of coils,...) necessary to build the signal model and initiate an hpi fit. 

### SignalModel
The signal model is utilized by the HpiFit algorithm to compute the HPI coil amplitudes. Since building and fitting the model is a task in itself, it was moved to its own class. Further, a struct `ModelParameters` was introduced. This struct is used to construct the signal model and is also one of the Inputs for the HpiFit class. The struct contains the information necessary for the SignalModel and therefore the HpiFitting. It contains the Hpi Coil frequencies, the line frequency, and whether to use the advanced model or not (should be removed in the future).

### HpiFit
The HpiFit class is now constructed using a SensorSet. Further, the number of input parameters was reduced, and the `findOrder ` and `fit ` functions were unified, since they do essentially the same, except 5 lines of code. Also, the results are now returned using the `HpiFitResult` struct.

Please find an example of the new usage of all these classes in the test_hpiFit Unit test. 

### ToDo
- [x] Check documentation of public functions. There is a lot of bad copy and paste right now
- [x] Check for Input Variable dimensions in HpiFit.fit()
- [x] rename ModelParameters to something like HpiModelParameters and make it an object
- [x] make SensorSet an object rather than struct (which is essentially an object in C++) to avoid direct access to important data.
- [x] make local variables `const` if necessary
- [x] add functionality to update the sensor set or create new HpiFit object when channels changed
- [ ] improve handling of single hpiFit / continuous hpiFit and findOrder
- [ ] add coil order function from @juangpc 

### Class Interaction Diagram
I hope the following makes the usage of the newly implemented classes a little transparent. Time flows from top to bottom.
Picture1


libraries/inverse/hpiFit/hpimodelparameters.h
```diff
@@ -0,0 +1,209 @@
...
+//=============================================================================================================
+
+inline int HpiModelParameters::iNHpiCoils() const
+{
+    return m_iNHpiCoils;
+}
+
+//=============================================================================================================
+
+inline int HpiModelParameters::iSampleFreq() const
```
LorenzE, line 166: why is this function not defined here like the other inlines?

libraries/inverse/hpiFit/sensorset.cpp
```diff
@@ -0,0 +1,151 @@
...
+SensorSet::SensorSet(const FwdCoilSet::SPtr pFwdSensorSet)
+{
+    if(pFwdSensorSet!=nullptr) {
+        convertFromFwdCoilSet(pFwdSensorSet);
+    }
+}
+
+//=============================================================================================================
+
+void SensorSet::convertFromFwdCoilSet(const QSharedPointer<FWDLIB::FwdCoilSet> pFwdSensorSet)
```
LorenzE, line 82: Can we refactor this to not be a shared pointer but const ref instead?

libraries/inverse/hpiFit/sensorset.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+    m_ncoils = pFwdSensorSet->ncoil;
+    m_np = pFwdSensorSet->coils[0]->np;
+    initMatrices(m_ncoils,m_np);
+
+    // get data froms Fwd Coilset
+    for(int i = 0; i < m_ncoils; i++){
```
LorenzE, line 89: Maybe use a range based for loop here. looks nice IMHO 
`for(auto* coil : pFwdSensorSet->coils){`

libraries/inverse/hpiFit/sensorset.cpp
LorenzE: Also the naming of the function might be confusing. How about `initFromFwdCoilSet`?

libraries/inverse/hpiFit/sensorset.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+        m_cosmag.block(i*m_np,0,m_np,3) = matCosmag;
+        m_rmag.block(i*m_np,0,m_np,3) = matRmag;
+    }
+    m_tra = MatrixXd::Identity(m_ncoils,m_ncoils);
+}
+
+//=============================================================================================================
+
+void SensorSet::initMatrices(int m_ncoils, int m_np)
```
LorenzE, line 119: do not use m_ for function variables. Also consider making the int const

libraries/inverse/hpiFit/sensorset.cpp
```diff
@@ -0,0 +1,151 @@
...
+SensorSetCreator::SensorSetCreator()
+{
+    QString qPath = QString(QCoreApplication::applicationDirPath() + "/resources/general/coilDefinitions/coil_def.dat");
+    m_pCoilDefinitions = FwdCoilSet::SPtr(FwdCoilSet::read_coil_defs(qPath));
+}
+
+//=============================================================================================================
+
+SensorSet SensorSetCreator::updateSensorSet(const QList<FIFFLIB::FiffChInfo>& channelList,
+                                            const Accuracy accuracy)
```
LorenzE, line 141: Can accuracy be a const ref?

libraries/inverse/hpiFit/signalmodel.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+MatrixXd SignalModel::fitData(const HpiModelParameters& hpiModelParameters, const MatrixXd& matData)
+{
+
+    if(checkEmpty(hpiModelParameters)) {
+        MatrixXd matTopo;
+        return matTopo;
+    }
+
+    bool bParametersChanged = m_modelParameters != hpiModelParameters;
```
LorenzE, line 87: consider marking variables as const if they do not change over the course of the function. Same goes for the rest of this PR.

libraries/inverse/hpiFit/signalmodel.cpp
```diff
@@ -0,0 +1,194 @@
...
+    bool bDimensionsChanged = m_iCurrentModelCols != matData.cols();
+
+    if(bDimensionsChanged || bParametersChanged) {
+        m_iCurrentModelCols = matData.cols();
+        m_modelParameters = hpiModelParameters;
+        selectModelAndCompute();
+    }
+
+    MatrixXd matTopo = m_matInverseSignalModel * matData.transpose();
+    return matTopo;
```
LorenzE, line 97: FYI: This is good example where copy elision aka NRVO will kick in :)

libraries/inverse/hpiFit/signalmodel.h
```diff
@@ -0,0 +1,158 @@
...
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ *
+ * @brief     SignalModel class declaration.
+ *
+ */
+
+#ifndef SignalModel_H
+#define SignalModel_H
```
LorenzE, line 36: All caps

libraries/inverse/hpiFit/signalmodel.h
```diff
@@ -0,0 +1,158 @@
...
+     * @param[in] hpiModelParameters   The HpiModelParameters.
+     * @param[in] matData           The data matrix.
+     *
+     * @return the fitted data
+     *
+     */
+    Eigen::MatrixXd fitData(const HpiModelParameters& hpiModelParameters,
+                            const Eigen::MatrixXd& matData);
+
+protected:
```
LorenzE, line 100: remove

LorenzE (commented):
Very cool! Thx @RDoerfel

LorenzE:
@RDoerfel Do you mind creating a rough draft of the class hierarchy and how they play together? If yes, you can just add the image to the PR description. I think it would benefit understating the new structure.

libraries/inverse/hpiFit/hpifit.cpp
```diff
@@ -81,345 +87,268 @@ using namespace FWDLIB;
...
-        if(pFiffInfo->dig[i].kind == FIFFV_POINT_HPI) {
-            iNumCoils++;
-            lHPIPoints.append(pFiffInfo->dig[i]);
-        }
-    }
+//=============================================================================================================
 
-    //Set coil frequencies
-    VectorXd vecCoilfreq(iNumCoils);
+void HPIFit::fit(const MatrixXd& matProjectedData,
```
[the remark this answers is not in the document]
RDoerfel: Was thinking about this as well. We need some info about the last fit, but we can store these as member variables. Might be a bit cleaner to use, but more to take care of within the class.

libraries/inverse/hpiFit/hpidataupdater.cpp
```diff
@@ -0,0 +1,218 @@
+//=============================================================================================================
+/**
+ * @file     hpidataupdater.cpp
+ * @author   Ruben Doerfel <doerfelruben@aol.com>
+ * @since    0.1.0
```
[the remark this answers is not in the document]
RDoerfel: Maybe we should try to adapt the template with each new release?

libraries/inverse/hpiFit/hpifit.cpp
```diff
@@ -543,61 +471,17 @@ Eigen::Matrix4d HPIFit::computeTransformation(Eigen::MatrixXd matNH, MatrixXd ma
...
-        sensors.cosmag.block(i*iNp,0,iNp,3) = matCosmag;
-        sensors.rmag.block(i*iNp,0,iNp,3) = matRmag;
-    }
-}
-
-//=============================================================================================================
-
 void HPIFit::storeHeadPosition(float fTime,
                                const Eigen::MatrixXf& transDevHead,
                                Eigen::MatrixXd& matPosition,
```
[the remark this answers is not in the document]
RDoerfel: storeHeadPosition appends matPosition with the current result. Maybe another name should be considered.

libraries/inverse/hpiFit/sensorset.cpp
RDoerfel: this happens when you do "Replace All"

libraries/inverse/hpiFit/hpifitdata.h
```diff
@@ -152,7 +153,7 @@ class INVERSESHARED_EXPORT HPIFitData
      * same output.
      */
     Eigen::MatrixXd compute_leadfield(const Eigen::MatrixXd& matPos,
-                                      const struct SensorSet& sensors);
+                                      const SensorSet sensors);
```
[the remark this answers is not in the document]
RDoerfel: I was on the hunt for a compilation error. Apparently, I forgot to clean up my mess :D

libraries/inverse/hpiFit/hpifit.h
```diff
@@ -259,64 +283,54 @@ class INVERSESHARED_EXPORT HPIFit
...
+    double objectTrans(const Eigen::MatrixXd matHeadCoil,
+                       const Eigen::MatrixXd matCoilsDev,
+                       const Eigen::MatrixXd matTrans);
 
-    Eigen::MatrixXd     m_matModel;         /**< The model that contains the sines/cosines for the hpi fit*/
-    bool                m_bDoFastFit;       /**< Do fast fit. */
-    QSharedPointer<FWDLIB::FwdCoilSet>  m_pCoilTemplate;    /**< */
-    QSharedPointer<FWDLIB::FwdCoilSet>  m_pCoilMeg;         /**< */
-    QVector<int>        m_vecFreqs;         /**< The frequencies for each coil in unknown order. */
+    SensorSet m_sensors{SensorSet()};            /**< The sensor struct that contains information about all sensors. */
```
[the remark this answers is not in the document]
RDoerfel: Because initially, I didn't have default constructors for these objects.

libraries/inverse/hpiFit/sensorset.cpp
RDoerfel: I get following error:
invalid range expression of type 'FWDLIB::FwdCoil **'; no viable 'begin' function available

libraries/inverse/hpiFit/sensorset.cpp
RDoerfel: > Can we refactor this to not be a shared pointer but const ref instead?

All members of FwdCoilSet return only pointers. So this would require a refactor of FwdCoilSet and not only my code.

libraries/inverse/hpiFit/signalmodel.cpp
RDoerfel: Are you talking about input variables?

libraries/inverse/hpiFit/hpidataupdater.cpp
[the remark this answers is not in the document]
LorenzE: good idea!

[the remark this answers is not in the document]
LorenzE: Do you want to open an issue for that?

libraries/inverse/hpiFit/hpidataupdater.cpp
```diff
@@ -0,0 +1,218 @@
+//=============================================================================================================
+/**
+ * @file     hpidataupdater.cpp
+ * @author   Ruben DÖrfel <doerfelruben@aol.com>
```
LorenzE, line 4: capital Ö. Check other files as well

libraries/inverse/hpiFit/sensorset.cpp
LorenzE: ok got it. `coils` is a double pointer and not an array or an object for which begin and end member functions. Please disregard my comment.

libraries/inverse/hpiFit/signalmodel.cpp
LorenzE: I mean local variables which are not changed after their initialisation and during their lifecycle 

`const bool bParametersChanged = m_modelParameters != hpiModelParameters;`

libraries/inverse/hpiFit/hpifit.cpp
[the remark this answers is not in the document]
RDoerfel: I think this also relates to one of my ToDos - "improve handling of single hpiFit / continuous hpiFit and findOrder". `HpiFitResult` has to be manually cleared outside of `HpiFit` for a single fit or fit with findOrder.

LorenzE:
@juangpc @gabrielbmotta Can you have a look at this PR as well please? We would like to go ahead and start a new round of testing with this version. Thanks!

Also, any ideas why the Codecov is failing?

gabrielbmotta:
Hi! The MEG is open on Friday for testing. We can merge as-is and then push any new changes after testing or test and incorporate changes to this PR. Whichever you prefer.

gabrielbmotta (approved):

LorenzE:
@gabrielbmotta Thanks for having a look at this. The PR was still WIP and I think Ruben wanted to push some last commits. No need to reopen though, @RDoerfel can you just create a subsequent PR?

PR #885, opened 2022-01-28
Neurofeedbackplugin

simonmarxgut:
Best regards from Tyrol
Simon Marxgut

UMIT TIROL

### DIFF
--- a/applications/mne_scan/libs/scDisp/realtimeneurofeedbackwidget.cpp
+++ b/applications/mne_scan/libs/scDisp/realtimeneurofeedbackwidget.cpp
@@ -1,0 +1,539 @@
+//=============================================================================================================
+/**
+ * @file     realtimeneurofeedbackwidget.cpp
+ * @author   Simon Marxgut <simon.marxgut@umit-tirol.at>
+ * @since    0.1.0
+ * @date     November, 2021
+ *
+ * @section  LICENSE
+ *
+ * Copyright (C) 2021, Simon Marxgut. All rights reserved.
+ *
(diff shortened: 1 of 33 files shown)


LorenzE:
@simonmarxgut Do you mind explaining the nature of these plugins? What kind of Neurofeedback paradigm did you implement and what kind of data acquisition did you use? Thx!

simonmarxgut:
Hey @LorenzE,

yes of course, sorry for the late reply. I had intended to write something but haven’t gotten around to it in the last few days.

The Neurofeedbackplugin is just for the visualisation for a neurofeedback generated by the Classifier Plugin, which is developed by Jinlong Dong (TU Ilmenau). Currently we are preparing another Plugin for the PR – Bandpower Plugin. Also this plugin could be used as input. But it could be that we split this into two separate plugins, one for continuous and one with discrete data.

There are three different types of output – Classifier, Frequency, and Balloon.

“Classifier” output allows visualisation of the classifications (Integers - Output Classifier) in the form of text or images.
“Frequency” output allows visualisation of different frequency components or frequency bands using sliders. There are some settings for the individual visualisation such as autoscaling.
The “Balloon” output is composed of two images – a background and an object with a white background. Originally, we used a landscape and a balloon, hence the name. The purpose for the white background is that the plugin removes all white pixels and puts the object on the background. Depending on the input of the Plugin, the Balloon starts to rise or sink. So, it is a Neurofeedback based on the principle of reward. Due to copyright concerns, no images are implemented by default.

During the measurement it is possible to change the channel selection, reset the autoscaling or set fixed minimum/maximum.

PR #893, opened 2022-05-13
[ENH] MNE-Scan: show gof as well

RDoerfel:
For a better understanding of whether fitting goes well and if the issue might be a particular coil, I added the Goodness of Fit (GoF ) next to the error in the `Quick Control` widget. 

In general, the error gives information about the transformation and the gof about the fit itself. So both should be presented next to each other


RDoerfel:
@juangpc this one is ready

juangpc:
Looking good. Thanks for your effort. I'll have a look.

gabrielbmotta (approved):

PR #898, opened 2022-08-31
[MAINT] Update windows runner for min version

gabrielbmotta:
Due to the [deprecation of the `windows-2016` runner](https://github.com/actions/runner-images/issues/5238), all runs that used `windows-2016` have been made to use `windows-2019` instead.


gabrielbmotta:
I've updated the branch protection rules to reflect the new win version.

juangpc (approved):

juangpc:
Good!

PR #902, opened 2022-10-31
[WIP] The Great CMake Migration

gabrielbmotta:
Currently have all the libraries, examples, and applications building with cmake. Using newer version of eigen that has cmake backend.

To do:

- Finish organizing project layout.
- Get all of the CI working.
- Get fftw integration with eigen working
- Implement all the build option definitions


gabrielbmotta:
On the question of naming authors on top of the file (above the license), what is the appropriate thing to do in this case? I see a lot of the .pro files have many many (many) authors; is the best thing to do to keep the same authorship in the new equivalent CMakeLists.txt file? Or since this is a new completely new file I should only include the names of the people who worked on the new CMakeLists file?

LorenzE:
> On the question of naming authors on top of the file (above the license), what is the appropriate thing to do in this case? I see a lot of the .pro files have many many (many) authors; is the best thing to do to keep the same authorship in the new equivalent CMakeLists.txt file? Or since this is a new completely new file I should only include the names of the people who worked on the new CMakeLists file?

I think it would be ok to remove all names and just add yours or leave names out of the files completley. Technically it should be enough to have the git history of the files to identify who worked on them?

juangpc:
At some point I gave it a thought. My 2cents: 
The whole preamble thing in header and implementation files is just an additional thing to take care of, while developing.
It adds zero value. 
It adds work load and importantly it distracts. **It gets in the way**
I would get rid of it all, in 99.9% of the files, if not all. 
Substitute it with something like the following or nothing at all: 
/*---------------------------------------------------------------------------------------------
 *  MNE-CPP Project. All rights reserved.
 *  Licensed under BSD-3 License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
Authors, emails, etc... is all stored in .git/ 
I understand that at some point it might had some meaning. But I now don't see any reason to use such long text.
I would only use something on header files.
The only reason why I haven't done this is because I'm wasn't sure if I was the only one.

I would also get rid of the //====.... rows. And the comment blocks announcing "header files" or "forward declarations"... they all get on the way. 

But I would understand that it might be just my opinion. 
My contribution: @LorenzE @gabrielbmotta your take.

gabrielbmotta:
> At some point I gave it a thought. My 2cents: The whole preamble thing in header and implementation files is just an additional thing to take care of, while developing. It adds zero value. It adds work load and importantly it distracts. **It gets in the way** I would get rid of it all, in 99.9% of the files, if not all. Substitute it with something like the following or nothing at all: /*---------------------------------------------------------------------------------------------
> 
>     * MNE-CPP Project. All rights reserved.
> 
>     * Licensed under BSD-3 License. See License.txt in the project root for license information.
>       _--------------------------------------------------------------------------------------------_/
>       Authors, emails, etc... is all stored in .git/
>       I understand that at some point it might had some meaning. But I now don't see any reason to use such long text.
>       I would only use something on header files.
>       The only reason why I haven't done this is because I'm wasn't sure if I was the only one.
> 
> 
> I would also get rid of the //====.... rows. And the comment blocks announcing "header files" or "forward declarations"... they all get on the way.
> 
> But I would understand that it might be just my opinion. My contribution: @LorenzE @gabrielbmotta your take.

I can definitely get behind the offloading the license to a single file and only doing a short header, and maybe adding a single text file with the name of all contributors to the project.

I do like the organization of having the `//=======` between definitions in the .cpp files, but i can see that maybe we do have too many in the beginning when saying what headers we define. I would not mind doing away with some of those and/or simplifying the comments so they don't take up three lines each.

gabrielbmotta:
If we do end up changing the headers, I would prefer doing it in a separate PR, if that is ok.

LorenzE:
> If we do end up changing the headers, I would prefer doing it in a separate PR, if that is ok.

Yes, that would be better. The changes in this PR are massive at it is :) Maybe let's talk about this in more detail in #908.

juangpc:
I think this is looking good. Close to merging state. @LorenzE  any comments?

juangpc (dismissed):
Yes

juangpc (approved):

lorenz-esch-snkeos:
@gabrielbmotta @juangpc Sry I did not read your comment. Great work! This merge really makes me happy :) Will it give a try when I have time :)

PR #912, opened 2023-02-24
Fix code coverage

gabrielbmotta:
Update code coverage script with correct paths and build with correct linker and compiler flags.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -453,8 +453,7 @@ jobs:
         modules: qtcharts
     - name: Configure and compile MNE-CPP
       run: |
-        cmake -DCMAKE_BUILD_TYPE="Release" -DWITH_CODE_COV=ON -B build -S src 
-        cmake --build build --parallel $(expr $(nproc --all) / 2)
+        ./tools/building/build_project.bat Release coverage
     - name: Run tests and upload results to Codecov
       env:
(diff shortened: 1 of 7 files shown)


juangpc:
Hi, 
added a bit of logic to the build script (in linux for now). 
out folder should also be able to store different build types.

.github/workflows/release.yml
```diff
@@ -453,7 +453,8 @@ jobs:
         modules: qtcharts
     - name: Configure and compile MNE-CPP
       run: |
-        ./tools/building/build_project.bat
+        cmake -DCMAKE_BUILD_TYPE="Release" -DWITH_CODE_COV=ON -B build -S src
```
juangpc, line 456: Release mode and coverage on?
Will --coverage override build mode and make it Debug?

tools/testing/test_all.bat
```diff
@@ -188,7 +188,7 @@ do
   fi
 
   if [ $RunCodeCoverage == "true" ]; then
-    find ./libraries -type f -name "*.cpp" -exec gcov {} \; &> /dev/null
+    find ./src/libraries -type f -name "*.cpp" -exec gcov {} \; &> /dev/null
```
juangpc, line 191: This is the part that needs to change to analyze different subfolders/projects in each report.

juangpc:
build script now windows compatible. (needs debug)

juangpc (approved):

PR #916, opened 2023-03-04
[ENH] Use more cpu

juangpc:
Hi, I've been trying to reduce compilation times. Noticed that with physical cores only ±50% of the cpu was used. Could you try this?


gabrielbmotta:
Looks like it will be faster using all available threads.

I got the following on the martinos system:
threads (32) = 3:44.73
cores (16) = 3:56.15

I got the following on my laptop:
threads (12) = 5:21.66
cores (6) = 5:57.88

gabrielbmotta (approved):

gabrielbmotta:
And if we turn off building tests and examples, the martinos compile time goes down by another thirty seconds, to 3:15.99

gabrielbmotta:
And ninja on my laptop I get it 20 seconds faster: 5:00.43 (still compiling examples and tests)

PR #927, opened 2023-03-14
[ENH] Use resources folder for sample data

juangpc:


juangpc:
@gabrielbmotta 
Wondering... Shouldn't it be out/Release.... so on?
https://github.com/mne-tools/mne-cpp/blob/8cb8beed8560fc9f227ffdc1b5fb7053270fb854/.github/workflows/release.yml#L570

juangpc:
Here in *Test the Build* (https://github.com/mne-tools/mne-cpp/blob/main/doc/gh-pages/pages/development/buildguide_cmake.md#test-the-build)
I would prefer to organize all references to sample data in one place only. What do you think?

gabrielbmotta:
> @gabrielbmotta Wondering... Shouldn't it be out/Release.... so on?
> 
> https://github.com/mne-tools/mne-cpp/blob/8cb8beed8560fc9f227ffdc1b5fb7053270fb854/.github/workflows/release.yml#L570

Yes. This also goes into a larger issue of the wasm build with cmake which still needs debugging.

juangpc:
so that I don't forget, let me post this here: 
todo, references to bin/ folder here
https://github.com/mne-tools/mne-cpp/blob/main/doc/gh-pages/pages/documentation/scan_prerecordeddata.md

juangpc:
> so that I don't forget, let me post this here: todo, references to bin/ folder here https://github.com/mne-tools/mne-cpp/blob/main/doc/gh-pages/pages/documentation/scan_prerecordeddata.md

done

juangpc:
Any build of any type links to the same folder through a full path sym link. 
You can move the out/Whatever folder around and everything should still work.
You can actually copy the data to the ```data``` folder in ```out/Whatever``` and you have a release you can deploy within that same platform.
Besides whatever bug cmake wants to throw at me, I think it is a reasonable solution this way. 
@gabrielbmotta what do you think?

juangpc:
Not sure what are ifw_installer, convention and wizards doing. How often they are used, etc... I've moved them to resources because they are not code or tooling scripts.

juangpc:
dropping this

PR #928, opened 2023-03-15
[FIX] Fix build for windows

juangpc:
In windows, dlls seem to be left compiled in the build folder when this RUNTIME_OUTPUT_DIRECTORY is not set.


juangpc:
Reorganization makes it more clear (to me) what is going on.
Don't mind deleting messages.

gabrielbmotta (approved):

PR #935, opened 2023-03-16
Add static build functionality to build script

gabrielbmotta:
Add static build functionality to build script and use it when building statically in the CI.


juangpc (dismissed):

juangpc:
Yes yes yes... not only static flag. enforcing what we do on the earth, up in the cloud.

juangpc (approved):

PR #936, opened 2023-03-17
[FIX] Fix passing static flag to cmake in buidl script

gabrielbmotta:
Fix build flag passed with `--D`. Should be `-D`.


juangpc (approved):

gabrielbmotta:
This PR's content is added in #937

PR #942, opened 2023-03-27
[FIX] Missing multiline backslash in build_project script

juangpc:
This comes even faster than expected.


juangpc:
This is tested and ready on my side.

gabrielbmotta (approved):

PR #950, opened 2023-04-18
WIP [ENH,MAINT] More build options

gabrielbmotta:
Add the following options to the build project script:

- [qt=<path>] - lets user specify a path to the desired qt installation to use to build the project.
- [all] - build all project components (libraries, applications, tests, examples)

Change cmake build flags to positive statements rather than double negatives (instead of setting DO_NOT_BUILD_X to OFF we now set BUILD_X to ON)
- Add some group build options for only building realtime or offline apps. (saves a bit of time if you need to rebuild from scratch)
- Turn off examples and tests by default - not too sure about this behavior, but it would save some time when developing. Maybe as an option instead? Thoughts?


tools/build_project.bat
```diff
@@ -339,7 +398,9 @@ doPrintHelp() {
   echo " "
   echo "[help]  - Print this help."
   echo "[mock]  - Show commands do not execute them."
+  echo "[all]   - Build entire project (libraries, applicatiojs, examples, tests)"
   echo "[clean] - Delete build and out folders for your configuration and exit."
+  echo "[clean-all] - Delete all build and out folders."
```
juangpc, line 403: ... and exit?

juangpc:
Thoughts/ideas on Specifics:
"clean-all" vs "rm -fr build out" . 
Qt version opt vs "-- -DQtDir=... "
"chill" vs "parallel=false"  (true by default)... 

We're reaching a point where the number of options is becoming  * almost * too broad. But I guess you probably agree on this, so you're adding all these options because... (my guess is that) it allows to interface always with the build script in the ci? Once we reach a point where all the ci workflows and our typical workflow already are able to use this build script, I think it is a good point to stop and do not add more options. After all there is this extra args thing after "--" which allows to directly interface with any option cmake might use.

juangpc (dismissed):

juangpc (dismissed):

juangpc (approved):

PR #967, opened 2023-11-06
[FIX] Revised connectivity examples, fixed some compilation errors with Qt6 and a missing include in the cmake files

johaenns:
1. disp3d.qrc was missing in the disp3D cmake
2. Including package OpenGLWidgets is required to properly compile on all platforms
3. Using std in applicagtionlogger.cpp led to compile problems on some platforms and is not necessary.
4. Revised connectivity examples to work properly.


juangpc:
Great. Thanks.
Whenever you are ready, we could merge.

johaenns:
Hi Juan,

if you have no further comments - as far as I'm concerned it's good to go.

Best, Johannes

PR #969, opened 2024-03-30
[FIX] Segfault in forward computation

johaenns:
Deleting the QSharedDataPointer causes a segfault when later dereferencing *m_eeg_forward_grad.data()/*m_meg_forward_grad.data() in the call of compute_forward_eeg/compute_forward_meg.

There might be a way to fix this while still deleting the (basically empty) object behind the shared pointer. However, as far as I can see we do not have a guarantee that the value of m_pSettings->compute_grad does not change, so if it would switch to true later on and for some reason calculatedFwd() gets called we would have a problem with the current code anyway.


johaenns:
Included the small bug fix in PR #970

PR #971, opened 2024-05-09
Bug fix for array comparison

ag-fieldline:
error: comparison between two arrays is deprecated in C++20 [-Werror=array-compare]


chdinh:
Fixed in v2.0. This PR appears to target an outdated branch; please rebase onto v2.0 (C++20 -Werror=array-compare) or provide the exact branch/toolchain you’re targeting.

PR #992, opened 2026-05-01
Remove codeql for staging

larsoner:
Looking at https://github.com/mne-tools/mne-cpp/actions/workflows/codeql.yml I see 9 runners in the last 2h that will each be tied up for about an hour

Screenshot 2026-05-01 at 10 21 02 AM

This PR aims to reduce this usage by just running CodeQL on `main`. I am not sure what the dev workflow is here exactly, but assuming you merge stuff from `staging` to `main` occassionally then this would help with the concurrency issues.

Feel free to close and fix a different way, but it would be good to get something like this in ASAP so that other repo workflows can continue!

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,9 +2,9 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "main", "staging" ]
+    branches: [ "main" ]
   pull_request:
-    branches: [ "main", "staging" ]
+    branches: [ "main" ]
   schedule:
     - cron: '28 15 * * 1'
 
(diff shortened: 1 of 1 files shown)


larsoner:
Ooof actually the runners are tied up for *over 3h* 😱

larsoner:
I am going to manually kill the CodeQL jobs in `staging` (except the latest commit), I hope that's okay @chdinh . Can restart them later if needed (esp. if a regression is seen)

chdinh:
@larsoner  Thanks for reporting and for looking into this. I agree it’s safe for CodeQL not to be triggered on staging; running it on main should be sufficient for now.

chdinh:
@larsoner, I have now introduced concurrency groups with cancel-in-progress: true. This should automatically cancel outdated runs when a newer run for the same branch/PR is triggered.

That should reduce the number of long-running jobs blocking the queue while still allowing us to include staging. What do you think?

larsoner:
> That should reduce the number of long-running jobs blocking the queue while still allowing us to include staging. What do you think?

We can try it! I'll keep an eye on the jobs and if they get backlogged consistently again we can revisit

PR #993, opened 2026-05-05
Segfault on gantry angle

ag-fieldline:
Segfault when opening FIF files saved from MNE Python. Gantry angle is saved as FIFFT_FLOAT by MNE Python and as FIFFT_INT by mne-cpp.


chdinh (approved):
Thanks for the contribution! This is a real-world interop bug — MNE-Python writes FIFF_GANTRY_ANGLE as FIFFT_FLOAT while mne-cpp expected FIFFT_INT, leading to a segfault. The type-guarded dispatch is exactly the right fix. Merging.

PR #994, opened 2026-05-05
init.sh fails due to incorrect index

ag-fieldline:
init.sh fails to run properly due to cmake_args index being incorrect. The CMAKE_PREFIX_PATH is actually index 6. Fixing it to not require an index at all.

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -368,17 +368,18 @@ if [[ ${DEPS_ONLY} -eq 1 ]]; then
     exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


chdinh (approved):
Thanks for the contribution! Two solid fixes: removing the brittle cmake_args index assumption in init.sh (pre-computing CMAKE_PREFIX_VALUE is much more robust), and correcting the Eigen bdcSvd template syntax in decoding_csp.cpp. Merging.

PR #1003, opened 2026-06-17
Loading FIFF would crash with empty raw-data directory

ag-fieldline:
Loading FIFF would crash with empty raw-data directory. Make sure it gracefully fails and that any indices being accessed are valid.

### DIFF
--- a/src/libraries/fiff/fiff_stream.cpp
+++ b/src/libraries/fiff/fiff_stream.cpp
@@ -1904,6 +1904,15 @@ bool FiffStream::setup_read_raw(QIODevice &p_IODevice,
     fiff_int_t first_samp = 0;
     fiff_int_t first_skip = 0;
     //
+    //  A malformed file may have a raw data block with an empty directory.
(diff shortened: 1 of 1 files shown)


chdinh (approved):
Approved as part of the v2.4.0 quality/robustness sweep.

The fix is correct and minimal: `FiffStream::setup_read_raw` indexed `dir[first]` without bounds checks, so a malformed FIFF whose raw-data block has an empty (or single-entry) directory would read past the end. Guarding `dir.isEmpty()` up front and `first < dir.size()` on the DATA_SKIP access resolves it. All CI is green.

Follow-up (I will push a small hardening commit to staging): the same routine derives per-buffer sample counts via `ent->size / (k * nchan)`, which divides by zero if a malformed measurement info reports 0 channels. I'll add an `nchan <= 0` guard alongside a regression test in `test_fiff_raw_io` covering the empty-directory case. Thanks @ag-fieldline for catching the original crash.

